PR #4, opened 2017-08-08
Build on host without merged '/usr'

d4s:

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -90,7 +90,7 @@ func charsToString(in []int8) string {
 	return string(s[0:i])
 }
 
-const initScript = `#!/usr/bin/busybox sh
+const initScript = `#!/bin/busybox sh
 
 busybox mount -t proc proc /proc
 busybox mount -t sysfs none /sys
--- a/machine.go
+++ b/machine.go
@@ -230,7 +230,7 @@ func (m *Machine) kernelRelease() string {
 		return release
 	}
 
-	files, err := ioutil.ReadDir("/usr/lib/modules")
+	files, err := ioutil.ReadDir("/lib/modules")
 	if err != nil {
 		log.Fatal(err)
 	}


machine.go
```diff
@@ -374,7 +378,12 @@ func (m *Machine) startup(command string, extracontent [][2]string) int {
 	w.WriteFile("/wrapper",
 		fmt.Sprintf(commandWrapper, command), 0755)
 
-	w.WriteFile("/init", initScript, 0755)
+	shebang := "#!/bin/busybox sh"
+	if mergedUsrSystem() {
+		shebang = "#!/usr/bin/busybox sh"
+	}
+	w.WriteFile("/init",
+		fmt.Sprintf("%s\n%s", shebang, initScript), 0755)
```
sjoerdsimons: Why this complexity, on a merged usr system /bin is a symlink to /usr/bin, so #!/bin/busybox should work in all cases afaik

machine.go
```diff
@@ -225,12 +224,17 @@ func (m *Machine) kernelRelease() string {
...
+	if _, err := os.Stat(path.Join(modpath, release)); err == nil {
 		return release
 	}
 
-	files, err := ioutil.ReadDir("/usr/lib/modules")
+	if mergedUsrSystem() {
+		modpath = "/usr/lib/modules"
+	}
+
+	files, err := ioutil.ReadDir(modpath)
```
sjoerdsimons: on a usr merged system /lib is a symlink to /usr/lib so doing a readdir on /lib/modules should always work without needing the extra determinations

machine.go
sjoerdsimons: Could you make the commit message explain the issue and solution a bit more.. Also for fakemachine "Build" isn't really a useful action, it would probably be start ? or correct initramfs build?

d4s: in fact initrd did not contain merged usr. It is a really quick fix, may be need to check initrd generation instead.

machine.go
d4s: generated initrd did not contain merged usr.

machine.go
d4s: Sure. Just to keep it here:

- initrd creation fails due non-existing /usr/lib/modules.
- initrd couldn't be started since /usr/bin/busybox absence.

Both problems are caused bu initrd generated without merged /usr

PR #6, opened 2017-10-24
New method addFstabEntry()

d4s:
AddFstabEntry() allows user to add own entries to '/etc/fstab' file
in fake machine. '/etc/fstab' is generated in common way for now.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -23,6 +23,7 @@ func mergedUsrSystem() bool {
 	return false
 }
 
+// Mapping host directory to mount point in fake machine
 type mountPoint struct {
 	hostDirectory    string
 	machineDirectory string
@@ -35,15 +36,26 @@ type image struct {
 }
 
 type Machine struct {
-	mounts []mountPoint
-	count  int
-	images []image
-	memory int
+	mounts       []mountPoint
+	fstabEntries []mountEntry
+	count        int
+	images       []image
+	memory       int
 
 	scratchsize int64
 	scratchpath string
 	scratchfile string
 	scratchdev  string
+}
+
+// Keeps entry to be added to '/etc/fstab' in fake machine
+type mountEntry struct {
+	fs_spec    string
+	fs_file    string
+	fs_vfstype string
+	fs_mntops  string
+	fs_freq    int
+	fs_passno  int
 }
 
 // Create a new machine object
@@ -160,6 +172,7 @@ SendSIGHUP=yes
 
 func (m *Machine) addStaticVolume(directory, label string) {
 	m.mounts = append(m.mounts, mountPoint{directory, directory, label})
+	m.addFstabEntry(label, directory, "9p", "trans=virtio,version=9p2000.L", 0, 0)
 }
 
 // AddVolumeAt mounts hostDirectory from the host at machineDirectory in the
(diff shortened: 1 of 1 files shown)


sjoerdsimons:
Could you add a note about the use-cases for this, i do wonder if you really want a raw AddFstabEntry function rather then say a CreateImageAt() or AddImageAt() functions ?

d4s:
In my particular case this PR allow to create external image outside of the library and mount prepared FS.
Generally we know nothing about attached image structure (partitions, filesystem type, options) but need to know how to mount prepared structure in fake machine. Since it is a common library it have sense to have a method for user to mount any device supported by linux kernel -- it could be external image prepared outside of fakemachine with partitions and/or strange filesystems, swap or even external network storage.

Usage of CreateImageAt() means the image must be prepared by library. This function and AddImageAt() also need to know everything about internals of the image and add proper fstab entries as well.

AddFstabEntry() adds enough flexibility to keep the logic in user's application without unneeded intelligence in the library:
- user prepare image (or use block device) and structure (partitions and filesystems) of the image outside of the library
- user instructs to attach the prepared image to VM with CreateImage(image, -1)
- user instructs what to mount and how to do it with AddFstabEntry() -- since user knows everything about internal structure of the image

sjoerdsimons:
You don't have to prepare things in the library; if you do your own prep and mounting rather then running it into fstab; fstab is more core of how fakemachine sets itself up (if it fails to mount  bits in fastab that's really annoying to debug as systemd will go into emergency mode)

If you want a single additional scratch image instead of /scratch ti seems useful to have that functionality in fakemachine itself in the first place. 

But i think my biggest question mark for me here is how you correlate the image files to device files inside the VM; The fstab tweaks will only help if you already prepped some known identifiers which doesn't seem very clean for most of the use-cases we care about

d4s:
> You don't have to prepare things in the library; if you do your own prep and mounting rather then running it into fstab;

hm... my idea was to start OS in Qemu allowing to deal systemd with images and block devices as I did it in [this commit](https://github.com/d4s/debos/commit/2b0b27752f73fab0fc67864bda6afc6d672e7e2c#diff-f0c7d8661440eb7e64c49f21358b3e16R62)

> fstab is more core of how fakemachine sets itself up (if it fails to mount bits in fstab that's really annoying to debug as systemd will go into emergency mode)

hm... just did a PoC with custom 'local-fs.target' there 'nFailure=emergency' is removed.
No emergency mode but didn't stopped at failure.

> But i think my biggest question mark for me here is how you correlate the image files to device files inside the VM

yep, this is the biggest problem. In case if you have something similar to AddFstabEntry() -- you just provide the string representing the mounting object -- "LABEL=...", "UUID=...", LVM volume name, external network storage (or something like '/dev/vda5' and cross the fingers ;))

sjoerdsimons:
Just did a quick PoC, but you can provide a serial ID for the virtio drive to qemu so you can correlate it back on the "other" side in a predictable way.

I'll try to create a PR for that tonight; but from that it's quite easy to label images in fakemachine for further usage inside the machine.

as also mentioned; i think this is feature that make sense directly in fake machine as well not just debos

d4s:
Reworked: now it is not exposed to the user (yet) but allows to create entrires to fstab in common way.

d4s:
Do not need the patch as it was a part of solution solved in a different way

PR #7, opened 2017-10-26
Prevent systemd to enter into emergency mode on mount error

d4s:
Add custom `local-fs.target` to prevent systemd to fall into
emergency mode in case of mount problems.
Removed option 'OnFailure=emergency.target' from custom version of
unit-file.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -268,8 +268,9 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) {
 	}
 }
 
-// Run creates the machine running the given command
-func (m *Machine) Run(command string) int {
+// Start the machine running the given command and adding the extra content to
+// the cpio. Extracontent is a list of {source, dest} tuples
+func (m *Machine) startup(command string, extracontent [][2]string) int {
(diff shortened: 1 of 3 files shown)


machine.go
```diff
@@ -148,6 +148,15 @@ IgnoreSIGPIPE=no
 SendSIGHUP=yes
 `
 
+const localfs = `
+[Unit]
+Description=Local File Systems
+Documentation=man:systemd.special(7)
+DefaultDependencies=no
+Conflicts=shutdown.target
+After=local-fs-pre.target
```
sjoerdsimons: That should really have a `FailureAction=poweroff-immediate` or somesuch; if mounting failes we want to bail from the fakemachine rather then muddling forwards with the invariants broken

sjoerdsimons: Actually it  would  be better not to overwrite local-fs.target but replace emergency.service by that calls poweroff immediately that would "fix" all cases where fakemachine would accidentally go into an emergency shell

d4s:
Do not need PR anymore since own systemd unit file is used as a final point for systemd

PR #14, opened 2017-11-17
Run go-debos on ubuntu

Thread974:
A couple patches are needed to run go-debos on ubuntu.
It is also needed to make the kernel user readable. `chmod go+r /boot/kernel`

### DIFF
--- a/cpio/writerhelper.go
+++ b/cpio/writerhelper.go
@@ -109,7 +109,7 @@ func (w *WriterHelper) CopyTree(path string) {
 		} else if info.Mode().IsRegular() {
 			w.CopyFile(p)
 		} else {
-			panic("No handled")
+			log.Printf("Ignored link: %s/%s", path, p);
 		}
 
 		return nil
--- a/machine.go
+++ b/machine.go
@@ -3,6 +3,7 @@ package fakemachine
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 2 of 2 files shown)


machine.go
```diff
@@ -323,8 +324,13 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) error {
 		if mergedUsrSystem() {
 			usrpath = "/usr/lib/modules"
 		}
-		if err := w.CopyFile(path.Join(usrpath, kernelRelease, v)); err != nil {
-			return err
+		log.Printf("Getting module from host %s", v)
```
sjoerdsimons: Drop the spurious debug output

machine.go
```diff
@@ -323,8 +324,13 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) error {
...
+		if (!strings.HasPrefix(v, "kernel/drivers/virtio")) {
+			if err := w.CopyFile(path.Join(usrpath, kernelRelease, v)); err != nil {
+				return err
+			}
+		} else {
+			log.Printf("Ignoring %s", v)
```
sjoerdsimons: Would be better to print a message about a module not being found and assuming it's build in (even better if we could determine that for sure, but tha'ts not always possible)

cpio/writerhelper.go
```diff
@@ -109,7 +109,7 @@ func (w *WriterHelper) CopyTree(path string) {
 		} else if info.Mode().IsRegular() {
 			w.CopyFile(p)
 		} else {
-			panic("No handled")
+			log.Printf("Ignored link: %s/%s", path, p);
```
sjoerdsimons: Stopping to panic is good; ignoring and moving forward is not.. Can you explain in your commit mesage what you're trying to solve here and why in that case it's reasonably to not copy?

Thread974: There used to by symbolic links to some open gl configuration in the ubuntu /etc.
I can't find them now, so it may no longer be required

PR #20, opened 2018-05-18
Dump diagnostics when the network fails to come up

sjoerdsimons:


machine.go
```diff
@@ -36,10 +36,10 @@ type image struct {
 }
 
 type Machine struct {
-	mounts []mountPoint
-	count  int
-	images []image
-	memory int
+	mounts  []mountPoint
```
d4s: not sure these blank spaces should be in current PR

machine.go
```diff
@@ -130,6 +129,17 @@ LinkLocalAddressing=no
...
+/lib/systemd/systemd-networkd-wait-online -q
+if [ $? != 0 ]; then
+  echo "WARNING: Network setup failed"
+  echo "== Journal =="
+  journalctl -a --no-pager
+  echo "== networkd =="
+  networkctl status
+  networkctl list
+  echo 1 > /run/fakemachine/result
+  exit
```
d4s: just exit? without errorcode?

sjoerdsimons: Yeah the exit code isn't pass through to the fakemachine running on the host just the result file

machine.go
sjoerdsimons: They're in a seperate commit; i guess it could be split out or done directly to master but i'll merge with rebase anyway so little harm

PR #24, opened 2018-06-06
main: Clarify the units used for -m, -s

smcv:
It wasn't previously obvious that `-m` expects a bare number of megabytes, but `-s` expects a unit suffix like `4G`.


sjoerdsimons:
Ugh should probably align that at some point. That's what you get from doing the quick simple thing of just passing a number onto qemu i guess :p

Issue #25, opened 2018-06-06
Virtual machine firmware/bootloader reconfigures tty


smcv:
Running qemu-system-x86_64 puts my terminal (gnome-terminal) in a weird mode where line wrapping and scrollback are disabled. I think this is caused by the firmware or bootloader.

autopkgtest-virt-qemu avoids this by only letting qemu-system-x86_64 write console messages to the invoking terminal if a special option (`--show-boot`) is given. fakemachine and debos could do similarly. For instance, without `--show-boot`, they could maybe send `/wrapper` output to `/dev/ttyS1` or `/dev/hvc1`, direct that character device to the invoking terminal, and use Unix sockets for `/dev/ttyS0`, `/dev/hvc0` and the qemu monitor? (`/dev/hvc0` is a bit special because systemd will use it as a virtual console by default, if it exists.)

PR #29, opened 2018-06-13
machine: Don't show boot/console messages by default

smcv:
In particular this avoids having the serial port support in the
BIOS/firmware reset the user's terminal into a weird state.

Signed-off-by: Simon McVittie <smcv@collabora.com>
Fixes: #25

### DIFF
--- a/cmd/fakemachine/main.go
+++ b/cmd/fakemachine/main.go
@@ -15,6 +15,7 @@ type Options struct {
 	Memory      int      `short:"m" long:"memory" description:"Amount of memory for the fakemachine in megabytes"`
 	CPUs        int      `short:"c" long:"cpus" description:"Number of CPUs for the fakemachine"`
 	ScratchSize string   `short:"s" long:"scratchsize" description:"On-disk scratch space size (with a unit suffix, e.g. 4G); if unset, memory backed scratch space is used"`
+	ShowBoot    bool     `long:"show-boot" description:"Show boot/console messages from the fakemachine"`
 }
 
 var options Options
@@ -78,6 +79,7 @@ func main() {
 	}
 
 	m := fakemachine.NewMachine()
+	m.SetShowBoot(options.ShowBoot)
 	SetupVolumes(m, options)
 	SetupImages(m, options)
 
--- a/machine.go
+++ b/machine.go
@@ -44,6 +44,7 @@ type Machine struct {
 	images  []image
 	memory  int
 	numcpus int
+	showBoot bool
 
 	scratchsize int64
 	scratchpath string
@@ -110,6 +111,7 @@ busybox mount -t proc proc /proc
 busybox mount -t sysfs none /sys
 
 busybox modprobe virtio_pci
+busybox modprobe virtio_console
 busybox modprobe 9pnet_virtio
 busybox modprobe 9p
 
@@ -166,6 +168,7 @@ ExecStopPost=/bin/sync
 ExecStopPost=/bin/systemctl poweroff -ff
 OnFailure=poweroff.target
 Type=idle
+TTYPath=%[1]s
 StandardInput=tty-force
 StandardOutput=inherit
 StandardError=inherit
@@ -263,6 +266,11 @@ func (m *Machine) SetNumCPUs(numcpus int) {
 	m.numcpus = numcpus
 }
 
+// SetShowBoot sets whether to show boot/console messages from the fakemachine.
+func (m *Machine) SetShowBoot(showBoot bool) {
+	m.showBoot = showBoot
+}
+
 // SetScratch sets the size and location of on-disk scratch space to allocate
 // (sparsely) for /scratch. If not set /scratch will be backed by memory. If
 // Path is "" then the working directory is used as a default storage location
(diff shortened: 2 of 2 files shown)


machine.go
```diff
@@ -501,11 +515,26 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
...
+			"-display", "none",
+			"-device", "virtio-serial",
+			"-chardev", "null,id=for-ttyS0",
+			"-chardev", "null,id=for-ttyS1",
+			"-chardev", "null,id=for-hvc0",
+			"-chardev", "stdio,id=for-hvc1",
+			"-device", "virtconsole,chardev=for-hvc0",
+			"-device", "virtconsole,chardev=for-hvc1",
+			"-serial", "chardev:for-ttyS0",
+			"-serial", "chardev:for-ttyS1")
```
sjoerdsimons: Would be nice to have some explanation of how this works as it's not entirely obvious to me (either in code comment or commit message).  I guess you're redirecting ttyS0/ttyS1 to /dev/null, and adding hvc consoles. But why ttyS1 as well, and similar why two hvc consoles rather then just one?

cmd/fakemachine/main.go
```diff
@@ -15,6 +15,7 @@ type Options struct {
 	Memory      int      `short:"m" long:"memory" description:"Amount of memory for the fakemachine in megabytes"`
 	CPUs        int      `short:"c" long:"cpus" description:"Number of CPUs for the fakemachine"`
 	ScratchSize string   `short:"s" long:"scratchsize" description:"On-disk scratch space size (with a unit suffix, e.g. 4G); if unset, memory backed scratch space is used"`
+	ShowBoot    bool     `long:"show-boot" description:"Show boot/console messages from the fakemachine"`
```
sjoerdsimons: The console output was never shown on purpose. Though it's useful for debugging, e.g. on some setups we've seen kernel panics in jenkins (due to nested virtualisation instabilities).

I do wonder whether it makes sense to have a specific show boot option or whether it should be a more general debug option which turns it on.

machine.go
```diff
@@ -50,7 +51,7 @@ type Machine struct {
 
 // Create a new machine object
 func NewMachine() (m *Machine) {
-	m = &Machine{memory: 2048, numcpus: runtime.NumCPU()}
+	m = &Machine{memory: 2048, numcpus: runtime.NumCPU(), showBoot: true}
```
sjoerdsimons: let showboot default to false

machine.go
smcv: The qemu command-line is basically a disaster area, with several layers of backwards compatibility where old special-case designs have become aliases for something newer and more general. `-nographic` is a shortcut for something resembling `-display none -chardev stdio,mux=on,id=foo -serial chardev:foo -mon chardev:foo` (possibly over-simplifying, I think `-nographic` might actually include a parallel port in the multiplexed character device too).

`-chardev null` isn't *exactly* `/dev/null`, but it's the equivalent in qemu code.

ttyS1 is unnecessary, I can drop that. I briefly thought I was going to have to use ttyS1, like autopkgtest-virt-qemu does, but then I got hvc1 working.

systemd special-cases the first virtualized console to put a getty on it automatically:

> "systemd will also spawn a login prompt on the first special VM console (that's /dev/hvc0, /dev/xvc0, /dev/hvsi0), if the system is run in a VM that provides these devices" —http://0pointer.de/blog/projects/serial-console.html

which is why I avoided using it; perhaps fakemachine is already blacklisting the relevant generator, but it seems best to avoid hvc0 for purposes that aren't actually the VM's interactive console.

cmd/fakemachine/main.go
smcv: I copied the name from `autopkgtest-virt-qemu`.

Would you prefer the boot/console messages etc. to go to a log file? That might be less annoying (and people can use `--debug-log /dev/stderr` if that's what they want).

machine.go
smcv: I was going for minimal changes, but I'd personally prefer the default to be false, so no objection to changing that :-)

machine.go
smcv: One thing we could maybe consider would be to use `console=hvc0`, so that there are three separate streams that you can have logged (or splatted onto stderr, as appropriate):

* ttyS0: BIOS messages corrupting your terminal, bootloader if used (but you're using direct kernel boot so that isn't applicable), and possibly the really early stages of Linux boot
* hvc0: systemd boot messages (I think these avoid corrupting the terminal state, although I'd need to test)
* hvc1: fakemachine/debos payload, aka what you actually wanted to see

cmd/fakemachine/main.go
smcv: I've left this as `--show-boot` until/unless we have a better name.

machine.go
smcv: For now I've changed this to:

* ttyS0: BIOS, bootloader if we used one, Linux, initramfs (?), systemd
* hvc0: reserved but unused (systemd might put a getty here, but if it does, we ignore it)
* hvc1: payload

machine.go
smcv: The default is now not showing boot messages.

machine.go
sjoerdsimons: To prevent systemd to put a getty there you can mask serial-getty@hvc0.service that's what we do for ttyS0

smcv: To confirm, you would prefer there to be only two separable streams: all the console stuff (from BIOS up to systemd), and the payload?

Issue #17, opened 2018-03-06
Fails to build on ARM


zumbi:
When attempting to build fakemachine for `armel` and `armhf` architectures, it fails to build with following error:
```
# github.com/go-debos/fakemachine
src/github.com/go-debos/fakemachine/machine.go:285:36: cannot use u.Release[:] (type []uint8) as type []int8 in argument to charsToString
```

Full build log can be found at
https://buildd.debian.org/status/fetch.php?pkg=golang-github-go-debos-fakemachine&arch=armhf&ver=0.0~git20180126.e307c2f-1&stamp=1519660834&raw=0

At a guess, u.Release[:] needs to be declared as the correct type or convert the value at the call.
I do wonder if `byte` might work better than `uint8` in this case.

smcv:
On the non-amd64 architectures where it builds successfully, it won't work anyway, due to #18.

obbardc:
fakemachine builds properly for arm64 now; we may add further architectures later if there is a need.

PR #30, opened 2018-06-14
[untested] machine: Add details of some more CPU architectures

### LINKED ISSUE #18: Assumes the host machine is always x86_64
Even if fakemachine is built for a non-x86_64 architecture, it hard-codes `qemu-system-x86_64` as the way to run the fake machine.

smcv:
fakemachine/debos are currently non-portable and should only be built for `amd64`, but they won't migrate to Debian testing until either they're built on various architectures where they previously built, or the archive administrators clean up the old binaries as requested in [#899312](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=899312) and [#899314](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=899314) (which they show no signs of doing).

For each supported CPU, we need to know the name of ld.so, the full name of libc.so, the qemu-system-* that emulates it, and what Go calls that CPU.

In particular this covers all the Go architectures mentioned in https://tracker.debian.org/media/packages/g/golang-defaults/control-21.105 except for `armel` (Go doesn't seem to distinguish between Debian `armel` and `armhf`, so I've assumed `armhf`). According to [the migration excuses](https://qa.debian.org/excuses.php?package=golang-github-go-debos-fakemachine), `armel` doesn't seem to be blocking migration right now anyway?

This fixes #18 if/when someone can test it on a non-x86_64 machine, most likely some sort of `arm64` like a Raspberry Pi 3.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -1,11 +1,12 @@
 // +build linux
-// +build amd64
+// +build 386 amd64 arm64 ppc64 ppc64le mips mipsle mips64 mips64le s390x
 
 package fakemachine
 
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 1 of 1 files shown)


smcv:
Or maybe it would be better to run `dpkg --print-architecture` and `switch` on the result of that, which would allow `armel` and `armhf` to be distinguished? (I don't know enough Go to implement this.)

smcv:
Ah, #17 means we don't currently need to care about 32-bit ARM for migration purposes, because fakemachine always failed to build from source there anyway.

machine.go
```diff
@@ -441,11 +442,67 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
...
-	w.CopyFile(prefix + "/bin/busybox")
 
-	/* Amd64 dynamic linker */
-	w.CopyFile("/lib64/ld-linux-x86-64.so.2")
+	qemu := ""
+
+	// Useful references:
+	// https://salsa.debian.org/go-team/compiler/golang/blob/golang-1.10/debian/helpers/goenv.sh
+	// https://sourceware.org/glibc/wiki/ABIList
+	switch runtime.GOARCH {
```
d4s: Please rework this to use table-like data structure instead of switch-case approach.
May be it is useful to define the single structure related to qemu here and use it in go-debos/debos/pull/91 ?

smcv:
Sorry, I don't know Go well enough to design data structures and library APIs in it. If someone who actually speaks the language can pick this up, that would be great!

sjoerdsimons:
Closing this one in favour of #157 which a baseline for more architectures ; However each architecture does need specific testing for qemu behaviour in particular, so you can't just plumb in a list of dynamic linkers like that and be done ;).. And really fakemachine is only really usable on architectures that support hw virtualisation

PR #31, opened 2018-06-15
Accept Qemu additional options

Lakshmipathi:
Pass custom qemu options while launching fakemachine.

Sample usage:
By default fakemachine starts Qemu with '-nographic', we
can modify it by passing below option,

qemuopts='-display sdl,gl=es -vga virtio -serial mon:stdio'

Signed-off-by: Lakshmipathi.G <lakshmipathi.ganapathi@collabora.co.uk>

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -44,6 +44,7 @@ type Machine struct {
 	images  []image
 	memory  int
 	numcpus int
+	showBoot bool
 
 	scratchsize int64
 	scratchpath string
@@ -111,6 +112,7 @@ busybox mount -t proc proc /proc
 busybox mount -t sysfs none /sys
 
 busybox modprobe virtio_pci
+busybox modprobe virtio_console
 busybox modprobe 9pnet_virtio
 busybox modprobe 9p
 
@@ -167,6 +169,7 @@ ExecStopPost=/bin/sync
 ExecStopPost=/bin/systemctl poweroff -ff
 OnFailure=poweroff.target
 Type=idle
+TTYPath=%[1]s
 StandardInput=tty-force
 StandardOutput=inherit
 StandardError=inherit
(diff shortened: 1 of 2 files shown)


machine.go
```diff
@@ -523,6 +529,15 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 				i, i, img.label))
 	}
 
+	if m.qemuopts != nil {
+		keyval := strings.Fields(m.qemuopts[0])
+		for _, v := range keyval {
+			qemuargs = append(qemuargs, v)
+		}
+	} else {
+		qemuargs = append(qemuargs, "-nographic")
```
tomeuv: isn't -nographic already added above?

Lakshmipathi: I moved `-nographic` from above to this else part. But didn't make that change while creating patch. Will make the change and update the patch. I was under the assumption` -nographic` and `-display` cannot be passed simultaneously. But qemu works fine even when both args are passed. So should I always keep `nographic` regardless of `-display` used or not?

smcv: See also #29 which switches from `-nographic` to `-display none`, and explicitly sets up the ttys separately (`-nographic` does multiple things, which can be confusing). It might be better to have an explicit `--display` option (mapped to qemu's `-display`, and defaulting to `none`) instead of passing it in via an "arbitrary qemu options" interface.

I wonder whether, longer-term, passing `-nodefaults` and setting everything up explicitly would be a better approach.

smcv:
> `qemuopts='... -serial mon:stdio'`

I think this is a sign of a problematic design. fakemachine itself should always set up the serial ports that it needs in order to do its job; a user should not be required to meet fakemachine's expectations "by hand".

#29 makes the serial port setup more explicit.

machine.go
Lakshmipathi: We initially discussed about using explicit qemu specific `-qemudisplay` option, later decided - we should accept arbitrary values since it will allow us to include more  options in future  without changing fake-machine parts.

smcv:
> For my purposes, -display egl-headless,gl=es is not enough, I also need -vga virtio.

Sure, I'm not saying you don't need the "escape hatch" of being able to add extra options, I'm just not so sure that it's the best way to pass in `-display`. (But perhaps qemu accepts multiple `-display` and ignores all except the last, I don't know.)

machine.go
Lakshmipathi: Okay, will remove `-nographic` and re-submit patch

Lakshmipathi:
Created new PR with above suggestions and resolved merge conflict. Its much more cleaner commit (don't have `merge` commits on pr) , so will close this as Duplicate of #33  If you have more suggestions let me know. thanks.

PR #33, opened 2018-07-09
Accept Qemu options from commandline

Lakshmipathi:
Pass custom qemu options while launching fakemachine.

Sample usage:
By default fakemachine starts Qemu with '-display none', we
can modify it by passing below option,

qemuopts='-display sdl,gl=es'

Signed-off-by: Lakshmipathi.G <lakshmipathi.ganapathi@collabora.co.uk>


smcv:
Does qemu (guarantee to) interpret `qemu-system-x86_64 -display none -display sdl` as being equivalent to `qemu-system-x86_64 -display sdl`?

Lakshmipathi:
Yes, I think so ` qemu-system-x86_64 -display none -display curses` overrides `none` and launches qemu with ncurses.

Lakshmipathi:
@sjoerd-ccu If this looks fine, can you merge it and close this PR?

sjoerdsimons:
@Lakshmipathi I'm rather on the fence about this one still; Is this something you're still interested in ?

Lakshmipathi:
@sjoerdsimons, This change looked useful at the time, but until now no-one seems to be blocked because this missing option. May be we can drop this change.

PR #34, opened 2018-08-17
machine.go: Add support to environment variables

petersenna:
Add Machine.SetEnviron() to configure environment variables that will be
used by the fakemachine runner systemd service.

Signed-off-by: Peter Senna Tschudin <peter.senna@collabora.com>

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -45,6 +45,7 @@ type Machine struct {
 	memory  int
 	numcpus int
 	showBoot bool
+	Environ []string
 
 	scratchsize int64
 	scratchpath string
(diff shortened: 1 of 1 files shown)


machine.go
```diff
@@ -151,6 +152,8 @@ echo Running '%[1]s'
 echo $? > /run/fakemachine/result
 `
 
+// The line 'Environment=%[2]s' is used for environemnt variables optionally
```
sjoerdsimons:
```suggestion
// The line 'Environment=%[2]s' is used for environment variables optionally
```

PR #36, opened 2019-01-28
machine.go: Fix kernel modules open failed by checking builtin modules

fdanis-oss:
This allows to use debos on Ubuntu (at least)

Signed-off-by: Frédéric Danis <frederic.danis@collabora.com>

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -352,10 +352,19 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) error {
 		"modules.builtin.bin",
 		"modules.devname"}
 
+	usrpath := "/lib/modules"
+	if mergedUsrSystem() {
+		usrpath = "/usr/lib/modules"
+	}
+	builtins := ""
+	data, err := ioutil.ReadFile(path.Join(usrpath, kernelRelease, "modules.builtin"))
+	if err == nil {
+		builtins = string(data)
+	}
+
 	for _, v := range modules {
-		usrpath := "/lib/modules"
-		if mergedUsrSystem() {
-			usrpath = "/usr/lib/modules"
+		if strings.Contains(builtins, v) {
+			continue
 		}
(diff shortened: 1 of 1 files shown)


simonschmeisser:
this seems to work fine on ubuntu 18.04 / neon user edition

machine.go
```diff
@@ -362,7 +362,20 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) error {
...
+		return err
+	}
+	builtins := string(data)
+
 	for _, v := range modules {
+		if strings.Contains(builtins, v) {
+			continue
+		}
 		usrpath := "/lib/modules"
 		if mergedUsrSystem() {
```
elboulangero: Dont you want to remove the `usrpath := ...` snippet here, since you moved it above already?

elboulangero: @fdanis-oss ^

machine.go
```diff
@@ -363,10 +363,19 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) error {
 		"modules.builtin.bin",
 		"modules.devname"}
 
+	usrpath := "/lib/modules"
+	if mergedUsrSystem() {
+		usrpath = "/usr/lib/modules"
+	}
+	data, err := ioutil.ReadFile(path.Join(usrpath, kernelRelease, "modules.builtin"))
```
sjoerdsimons: On debian systems modules.builtin doesn't exist at least on the 4.19 kernels it does however exist on  the 5.3 packaging. So the code should handle that properly (e.g. assume nothing is builtin like its' done now)

andrewshadura: In any case, we’re unconditionally copying `modules.builtin` below, so the case when it doesn’t exist has already *not* been covered.

obbardc:
Reworked and merged in #47, hence closing :-)

PR #41, opened 2019-08-08
Run fakemachine on ArchLinux

elboulangero:
This merge request allows fakemachine to run on ArchLinux.

Most of the changes are about finding files in different locations on the host, in order to create the initrd.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -469,11 +469,11 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 		w.WriteDirectory("/lib", 0755)
 	}
 
-	prefix := ""
 	if mergedUsrSystem() {
-		prefix = "/usr"
+		w.CopyFile("/usr/bin/busybox")
+	} else {
+		w.CopyFile("/bin/busybox")
 	}
-	w.CopyFile(prefix + "/bin/busybox")
 
 	/* Amd64 dynamic linker */
 	w.CopyFile("/lib64/ld-linux-x86-64.so.2")
--- a/machine.go
+++ b/machine.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
@@ -25,6 +26,20 @@ func mergedUsrSystem() bool {
 	}
 
 	return false
+}
+
+// Evaluate any symbolic link, then return the path's directory. Returns an
+// absolute path. Think of it as realpath(1) + dirname(1) in bash.
+func realDir(path string) (string, error) {
+	var p string
+	var err error
+	if p, err = filepath.Abs(path); err != nil {
+		return "", err
+	}
+	if p, err = filepath.EvalSymlinks(p); err != nil {
+		return "", err
+	}
+	return filepath.Dir(p), nil
 }
 
 type mountPoint struct {
@@ -458,12 +473,18 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 	if mergedUsrSystem() {
 		prefix = "/usr"
 	}
-	w.CopyFile(prefix + "/lib/x86_64-linux-gnu/libresolv.so.2")
-	w.CopyFile(prefix + "/lib/x86_64-linux-gnu/libc.so.6")
 	w.CopyFile(prefix + "/bin/busybox")
 
 	/* Amd64 dynamic linker */
 	w.CopyFile("/lib64/ld-linux-x86-64.so.2")
+
+	/* C libraries */
+	libraryDir, err := realDir("/lib64/ld-linux-x86-64.so.2")
+	if err != nil {
+		return -1, err
+	}
+	w.CopyFile(libraryDir + "/libc.so.6")
+	w.CopyFile(libraryDir + "/libresolv.so.2")
 
 	w.WriteCharDevice("/dev/console", 5, 1, 0700)
 
--- a/machine.go
+++ b/machine.go
@@ -4,6 +4,7 @@
 package fakemachine
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -351,6 +352,49 @@ func (m *Machine) kernelRelease() (string, error) {
 	return (files[len(files)-1]).Name(), nil
 }
 
+func (m *Machine) kernelPath(kernelRelease string) (string, error) {
+	kernelDir := "/boot"
+
+	/* First, try to find a kernel with a well-known name */
+	path := filepath.Join(kernelDir, "vmlinuz-" + kernelRelease)
+	if _, err := os.Stat(path); err == nil {
+		return path, nil
+	}
+
+	/* Otherwise, inspect each kernel installed, and look for the release
+	 * string straight in the binary. Not pretty, but it works. */
+	needle := kernelRelease
+	if !strings.HasSuffix(needle, " ") {
+		needle += " "
+	}
+
+	files, err := ioutil.ReadDir(kernelDir)
+	if err != nil {
+		return "", err
+	}
+
+	for _, f := range files {
+		if !strings.HasPrefix(f.Name(), "vmlinuz-") || f.IsDir() {
+			continue
+		}
+
+		path := filepath.Join(kernelDir, f.Name())
+		buf, err := ioutil.ReadFile(path)
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "Failed to read kernel:", err)
+			continue
+		}
+
+		if !bytes.Contains(buf, []byte(needle)) {
+			continue
+		}
+
+		return path, nil
+	}
+
+	return "", fmt.Errorf("No kernel found for release %s", kernelRelease)
+}
+
 func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) error {
 	kernelRelease, err := m.kernelRelease()
 	if err != nil {
@@ -544,6 +588,10 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 	if err != nil {
 		return -1, err
 	}
+	KernelPath, err := m.kernelPath(kernelRelease)
+	if err != nil {
+		return -1, err
+	}
 	memory := fmt.Sprintf("%d", m.memory)
 	numcpus := fmt.Sprintf("%d", m.numcpus)
 	qemuargs := []string{"qemu-system-x86_64",
@@ -551,7 +599,7 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 		"-smp", numcpus,
 		"-m", memory,
 		"-enable-kvm",
-		"-kernel", "/boot/vmlinuz-" + kernelRelease,
+		"-kernel", KernelPath,
 		"-initrd", InitrdPath,
 		"-display", "none",
 		"-no-reboot"}


elboulangero:
Is there anyone willing to feedback on this PR?

gasinvein:
Hey there. I've made and attempt to package debos for Fedora and stumbled upon this paths issue, too.
It would be nice if fakemachine could be more distro-agnostic. Maybe detect host OS via `/etc/os-release` and pick the right paths for needed files based on per-distro-family set?

elboulangero:
@gasinvein Did you try this patch set on a Fedora machine?

gasinvein:
@elboulangero No, but as far as I can read Go it won't help much: Fedora (and openSUSE at least) keeps 64-bit libs in `/usr/lib64`, `/usr/lib` is for 32-bit ones.

machine.go
```diff
@@ -374,8 +374,10 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) error {
 		if mergedUsrSystem() {
 			usrpath = "/usr/lib/modules"
 		}
-		if err := w.CopyFile(path.Join(usrpath, kernelRelease, v)); err != nil {
-			return err
+		if _, err := os.Stat(path.Join(usrpath, kernelRelease, v)); err == nil {
+			w.CopyFile(path.Join(usrpath, kernelRelease, v))
+		} else {
+			w.CopyFile(path.Join(usrpath, kernelRelease, v + ".xz"))
```
sjoerdsimons: Removing the error here can cause hard to debug issues if some modules are missing so i'd rather not do that. In general fakemachine should be better at error checking, it being a bit limited was that it started out as a PoC that turned into something actually used. (I've long hoped to get time to refactor the code quite a bit but never got aroudn to it)

machine.go
```diff
@@ -538,14 +538,18 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 	if err != nil {
 		return -1, err
 	}
+	KernelPath := "/boot/vmlinuz-" + kernelRelease
+	if _, err := os.Stat(KernelPath); err != nil {
+		KernelPath = "/boot/vmlinuz-linux"
+	}
```
sjoerdsimons: I'm not against it but it seems very haphazard. How can we be sure the modules in the initramfs match the kernel version ?

sjoerdsimons: Does Arch not suppot multiple parallel installed kernels or ?

machine.go
```diff
@@ -464,8 +464,17 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
...
+	if _, err := os.Stat(prefix + "/lib/x86_64-linux-gnu/libresolv.so.2"); err == nil {
+		w.CopyFile(prefix + "/lib/x86_64-linux-gnu/libresolv.so.2")
+	} else {
+		w.CopyFile(prefix + "/lib/libresolv.so.2")
+	}
+	if _, err := os.Stat(prefix + "/lib/x86_64-linux-gnu/libc.so.6"); err == nil {
+		w.CopyFile(prefix + "/lib/x86_64-linux-gnu/libc.so.6")
+	} else {
+		w.CopyFile(prefix + "/lib/libc.so.6")
+	}
```
sjoerdsimons: this could be a lot nicer if we had a list of reasonable prefixes (and probably a list of files to copy) or a dedicated function to find the file in the various paths.. For the original code what wasn't done as it was just two CopyFile calls but this change adds a lot of extra noise

sjoerdsimons:
Fwiw the patches that i didn't comment on looked fine to me and there is no real interdependency so you could submit those for merge seperatly (github doesn't provide nice ui to pick an choose)

machine.go
gasinvein: Can we just `readlink` it? AFAIK `/lib/libc.so.6` and `/lib64/ld-linux-x86-64.so.2` are well-known paths that are always there.

sjoerdsimons: `libc.so.6` is not a well-known path; it's just another library.  The linker is a well known-path per architecture as defined by  https://sourceware.org/glibc/wiki/ABIList . 

On my debian system:
```
rwxrwxrwx 1 root root 32 Oct 31 23:31 /lib64/ld-linux-x86-64.so.2 -> /lib/x86_64-linux-gnu/ld-2.29.so
```

So i guess one could readlink it and work out a reasonable prefix from that; Though i don't know if on other distributions it's also a symlink to the same prefix as where all other libraries live ?

gasinvein: > `libc.so.6` is not a well-known path; it's just another library

True, sorry, confused it with `ld-linux.so.2`.

> it's also a symlink to the same prefix as where all other libraries live ?

I guess it's very likely to be this way.

gasinvein: Maybe we could use `ldconfig -p` or something similar to find libraries?

elboulangero:
> Fwiw the patches that i didn't comment on looked fine to me and there is no real interdependency so you could submit those for merge seperatly (github doesn't provide nice ui to pick an choose)

I submitted https://github.com/go-debos/fakemachine/pull/44 for the 3 patches that were fine (I took this chance to improve the commit messages)

machine.go
elboulangero: So Arch does support parallel kernel installations, but not in the sense "different versions", but more "different packages, that happen to be different versions". The idea of including the version in the filenames and paths is discussed at: https://bugs.archlinux.org/task/16702. Unfortunately the discussion has been ongoing for 10 years.

The two more common packages that you might want to install are `linux` and `linux-lts` (from the `core` repository), and there is also `linux-zen` and `linux-hardened` from the `extra` repository.

If I install them all, here's what I get:

```
# ls -l /lib/modules
total 16
drwxr-xr-x 3 root root 4096 Dec 10 07:31 4.19.88-1-lts
drwxr-xr-x 3 root root 4096 Dec 10 07:31 5.3.15.a-1-hardened
drwxr-xr-x 3 root root 4096 Dec 10 07:33 5.4.2-arch1-1
drwxr-xr-x 3 root root 4096 Dec 10 07:28 5.4.2-zen1-1-zen

# ls -l /boot/
total 162528
-rw-r--r-- 1 root root 26461234 Dec 10 07:33 initramfs-linux-fallback.img
-rw-r--r-- 1 root root 27048174 Dec 10 07:31 initramfs-linux-hardened-fallback.img
-rw-r--r-- 1 root root  8465635 Dec 10 07:31 initramfs-linux-hardened.img
-rw-r--r-- 1 root root  8411613 Dec 10 07:33 initramfs-linux.img
-rw-r--r-- 1 root root 25952625 Dec 10 07:31 initramfs-linux-lts-fallback.img
-rw-r--r-- 1 root root  8422514 Dec 10 07:31 initramfs-linux-lts.img
-rw-r--r-- 1 root root 27433250 Dec 10 07:28 initramfs-linux-zen-fallback.img
-rw-r--r-- 1 root root  8455012 Dec 10 07:28 initramfs-linux-zen.img
-rw-r--r-- 1 root root  6404480 Dec 10 07:33 vmlinuz-linux
-rw-r--r-- 1 root root  6527360 Dec 10 07:31 vmlinuz-linux-hardened
-rw-r--r-- 1 root root  5896064 Dec 10 07:31 vmlinuz-linux-lts
-rw-r--r-- 1 root root  6924672 Dec 10 07:28 vmlinuz-linux-zen
```

At a quick glance, it's impossible to reliably map what's in `/usr/lib/modules` with the kernels in `/boot` in a generic way. Unless we want to make debos "archlinux aware" and start grepping around for keywords like `lts`, `hardened` and so on, but I'd prefer to avoid that.

I think a better solution would be to scan for files `/boot/vmlinux-*`, and get to know more about these files. For example, with `file(1)` we can actually know the kernel version:

```
$ file /boot/vmlinuz-linux*
/boot/vmlinuz-linux:          Linux kernel x86 boot executable bzImage, version 5.4.2-arch1-1 (linux@archlinux) #1 SMP PREEMPT Thu, 05 Dec 2019 12:29:40 +0000, RO-rootFS, swap_dev 0x6, Normal VGA
/boot/vmlinuz-linux-hardened: Linux kernel x86 boot executable bzImage, version 5.3.15.a-1-hardened (linux-hardened@archlinux) #1 SMP PREEMPT Thu, 05 Dec 2019 22:09:18 +0000, RO-rootFS, swap_dev 0x6, Normal VGA
/boot/vmlinuz-linux-lts:      Linux kernel x86 boot executable bzImage, version 4.19.88-1-lts (linux-lts@archlinux) #1 SMP Thu, 05 Dec 2019 12:30:39 +0000, RO-rootFS, swap_dev 0x5, Normal VGA
/boot/vmlinuz-linux-zen:      Linux kernel x86 boot executable bzImage, version 5.4.2-zen1-1-zen (linux-zen@archlinux) #1 ZEN SMP PREEMPT Thu, 05 Dec 2019 12:30:04 +0000, RO-rootFS, swap_dev 0x6, Normal VGA
```

I wonder if we could either use file and parse the output, or directly open the vmlinuz file and look straight in the right spot to know the version.

elboulangero: FWIW, ArchLinux's derivative Manjaro is a bit more similar to Debian, in the sense that it provides different packages to install different specific version of the kernel, for example `linux419` for `4.19`, `linux53` for `5.3`. They also provide packages for realtime kernel, eg. `linux419-rt`.

However, once again you can't directly use the kernel version in `/usr/lib/modules` to find the kernel in `/boot`:

```
# ls -l /usr/lib/modules
drwxr-xr-x 4 root root 4096 Dec 10 08:36 4.19.72-rt26-MANJARO
drwxr-xr-x 4 root root 4096 Dec 10 08:31 4.19.85-1-MANJARO
drwxr-xr-x 4 root root 4096 Dec 10 08:31 5.3.12-1-MANJARO

# ls -l /boot
-rw-r--r-- 1 root root 30711913 Dec 10 08:36 initramfs-4.19-rt-x86_64-fallback.img
-rw-r--r-- 1 root root  8161244 Dec 10 08:36 initramfs-4.19-rt-x86_64.img
-rw-r--r-- 1 root root 30646818 Dec 10 08:31 initramfs-4.19-x86_64-fallback.img
-rw-r--r-- 1 root root  8159883 Dec 10 08:31 initramfs-4.19-x86_64.img
-rw-r--r-- 1 root root 31060316 Dec 10 08:32 initramfs-5.3-x86_64-fallback.img
-rw-r--r-- 1 root root  8168645 Dec 10 08:31 initramfs-5.3-x86_64.img
-rw-r--r-- 1 root root       27 Nov 11 19:00 linux-419-rt-x86_64.kver
-rw-r--r-- 1 root root       22 Nov 21 10:36 linux419-x86_64.kver
-rw-r--r-- 1 root root       21 Nov 21 10:55 linux53-x86_64.kver
-rw-r--r-- 1 root root  5924736 Dec 10 08:36 vmlinuz-4.19-rt-x86_64
-rw-r--r-- 1 root root  5957504 Dec 10 08:31 vmlinuz-4.19-x86_64
-rw-r--r-- 1 root root  6343040 Dec 10 08:31 vmlinuz-5.3-x86_64
```

Once again, using `file` (or `strings`) to find the kernel version would work.

@sjoerd Is it acceptable? `file`, `string` or something else? Or do you think of a better solution?

machine.go
elboulangero: > So i guess one could readlink it and work out a reasonable prefix from that; Though i don't know if on other distributions it's also a symlink to the same prefix as where all other libraries live ?

It works for ArchLinux and Manjaro (where `/lib`, `/lib64` and `/usr/lib64` are all symlinks to `/usr/lib`).

Just to be sure, I had a look at Ubuntu 18.04 LTS, it's also the case

> I guess it's very likely to be this way.

So you mean it's also true for Fedora.

sjoerdsimons:
```
$ docker run -it fedora ls -l /lib64/ld-linux-x86-64.so.2
lrwxrwxrwx 1 root root 10 Jun  6  2019 /lib64/ld-linux-x86-64.so.2 -> ld-2.29.so
$ docker run -it fedora ls -l /lib64/libc.so.6
lrwxrwxrwx 1 root root 12 Jun  6  2019 /lib64/libc.so.6 -> libc-2.29.so
```
So yeah i would expect the "what directory does the linker live" to be a reasonableish guess for most libraries.

machine.go
sjoerdsimons: Wonder if there is a golang library that can help here (to get the string out) rather then hacking it via shell commands; But yeah doing that would be a good way going forward

machine.go
```diff
@@ -454,17 +513,23 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 		w.WriteDirectory("/lib", 0755)
 	}
 
-	prefix := ""
 	if mergedUsrSystem() {
-		prefix = "/usr"
+		w.CopyFile("/usr/bin/busybox")
```
gasinvein: On Fedora busybox is under `/usr/sbin/busybox`. Maybe search `PATH` for it instead of checking only `/bin` and `/usr/bin`?

elboulangero: Good point, thanks :)

elboulangero:
@sjoerd-ccu I submitted https://github.com/go-debos/fakemachine/pull/45, in order to rework modules handling. It includes the commit to support xz-compressed modules, which is now out of this PR.

elboulangero:
@sjoerd-ccu I updated this PR with changes as discussed, this is now ready to review.

machine.go
elboulangero: I implemented it this way in the updated PR.

machine.go
elboulangero: Couldn't find a golang library, so I wrote it myself. It's actually trivial to open a file and find a string.

machine.go
elboulangero: OK, i reworked that and it's now part of https://github.com/go-debos/fakemachine/pull/45

gasinvein:
Any update on this?

machine.go
obbardc: opened that as a separate issue ; it doesn't block this MR: https://github.com/go-debos/fakemachine/issues/48

santoshbit20071:
A new PR is opened with more fixes/improvements which supersedes this PR here : https://github.com/go-debos/fakemachine/pull/62

sjoerdsimons:
Closing in favour of #62

PR #43, opened 2019-09-27
Add support for per-volume virtfs and mount options

branen:
This change adds support for adding additional options and overriding default options for the virtfs arguments passed to QEMU and the fstab entries placed in the fakemachine guest.  It is fully backwards-compatible with the existing CLI and library interfaces; existing code will build against it and existing scripts will run against it without any change in behavior. 

Notably, this change allows volumes to be mounted with virtfs `security_model=mapped-file`, which in turn allows fakemachine to address edge cases that fakeroot cannot.

### DIFF
--- a/cmd/fakemachine/main.go
+++ b/cmd/fakemachine/main.go
@@ -21,15 +21,35 @@ type Options struct {
 var options Options
 var parser = flags.NewParser(&options, flags.Default)
 
+func parseVolumeOptions(optionString string) (mpo []fakemachine.MountPointOption) {
+	mpo = make([]fakemachine.MountPointOption, 0)
+	opts := strings.Split(optionString, ",")
+	for _, opt := range opts {
+		parts := strings.SplitN(opt, ".", 2)
+		switch parts[0] {
+		case "virtfs":
+			mpo = append(mpo, fakemachine.VirtfsOption(parts[1]))
+		case "mount":
+			mpo = append(mpo, fakemachine.MountOption(parts[1]))
+		default:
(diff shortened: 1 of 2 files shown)


cmd/fakemachine/main.go
```diff
@@ -21,15 +21,35 @@ type Options struct {
...
+func parseVolumeOptions(optionString string) (mpo []fakemachine.MountPointOption) {
+	mpo = make([]fakemachine.MountPointOption, 0)
+	opts := strings.Split(optionString, ",")
+	for _, opt := range opts {
+		parts := strings.SplitN(opt, ".", 2)
+		switch parts[0] {
+		case "virtfs":
+			mpo = append(mpo, fakemachine.VirtfsOption(parts[1]))
+		case "mount":
+			mpo = append(mpo, fakemachine.MountOption(parts[1]))
```
sjoerdsimons: I had to wrap my head a few times to work out how this worked ;). 

I'd change this to just add another fucntion (AddVolumeAtWithOptions or something with a nicer name :p) to pass on the options directly rather then parsing the string into a list of funcs which then get executed to add themselves to the mountpoint :)..

Bonus point for a creative solution though.

obbardc:
Have you got any further information on:

> Notably, this change allows volumes to be mounted with virtfs security_model=mapped-file, which in turn allows fakemachine to address edge cases that fakeroot cannot.

sjoerdsimons:
Closing as the author is unresponse and it's not entirely clear why this is needed

PR #45, opened 2019-12-11
Improve modules support

elboulangero:
This pull request is a rework of https://github.com/go-debos/fakemachine/pull/36, hence deprecates it.

It brings two additions in the way fakemachine handles kernel module:
1. Support the file `modules.builtin` (if present), so that when some of the modules that fakemachine needs are built in the kernel, fakemachine doesn't fail trying to copy it. This is needed in latest version of Ubuntu and ArchLinux, where the `virtio` modules are now builtin.
2. Support for xz-compressed modules. This is needed for ArchLinux.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -106,6 +106,15 @@ func charsToString(in []int8) string {
 	return string(s[0:i])
 }
 
+func stringInSlice(a string, list []string) bool {
+	for _, b := range list {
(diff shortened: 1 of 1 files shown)


obbardc:
Reworked and merged in #47, hence closing :-)

Issue #13, opened 2017-11-08
Add configurable kernel modules list


d4s:
Some distros have compiled-in kernel modules like ubuntu: `CONFIG_VIRTIO=y`

Fakemachine crashes with following log:
```
2017/11/08 16:30:47 open failed: /lib/modules/4.10.0-38-generic/kernel/drivers/virtio/virtio.ko - open /lib/modules/4.10.0-38-generic/kernel/drivers/virtio/virtio.ko: no such file or directory
--- FAIL: TestSuccessfullCommand (0.00s)
panic: open failed: /lib/modules/4.10.0-38-generic/kernel/drivers/virtio/virtio.ko - open /lib/modules/4.10.0-38-generic/kernel/drivers/virtio/virtio.ko: no such file or directory [recovered]
    panic: open failed: /lib/modules/4.10.0-38-generic/kernel/drivers/virtio/virtio.ko - open /lib/modules/4.10.0-38-generic/kernel/drivers/virtio/virtio.ko: no such file or directory

goroutine 5 [running]:
panic(0x64b6a0, 0xc4200f52e0)
    /usr/lib/go-1.7/src/runtime/panic.go:500 +0x1a1
testing.tRunner.func1(0xc420094180)
    /usr/lib/go-1.7/src/testing/testing.go:579 +0x25d
panic(0x64b6a0, 0xc4200f52e0)
    /usr/lib/go-1.7/src/runtime/panic.go:458 +0x243
log.Panicf(0x6ac7ed, 0x14, 0xc420053800, 0x2, 0x2)
    /usr/lib/go-1.7/src/log/log.go:327 +0xe3
github.com/go-debos/fakemachine/cpio.(*WriterHelper).CopyFileTo(0xc42000d5b0, 0xc420016d00, 0x3e, 0xc420016d00, 0x3e, 0x0, 0x0)
    /home/fredo/Documents/bos0009/go/src/github.com/go-debos/fakemachine/cpio/writerhelper.go:126 +0x1b1
github.com/go-debos/fakemachine/cpio.(*WriterHelper).CopyFile(0xc42000d5b0, 0xc420016d00, 0x3e, 0xc420016d00, 0x3e)
    /home/fredo/Documents/bos0009/go/src/github.com/go-debos/fakemachine/cpio/writerhelper.go:150 +0x49
github.com/go-debos/fakemachine.(*Machine).writerKernelModules(0xc420096400, 0xc42000d5b0, 0x2a, 0x6b64fc)
[...]

PR #47, opened 2020-02-27
Don’t attempt to copy built-in modules

andrewshadura:
Several distributions (including Ubuntu) ship some of the modules we rely on compiled into the kernel. Since these modules are listed in `modules.builtin`, it’s very easy to detect and skip them.

Signed-off-by: Andrej Shadura <andrew.shadura@collabora.co.uk>

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -352,12 +352,13 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) error {
 		"modules.builtin.bin",
 		"modules.devname"}
 
+	moddir := "/lib/modules"
+	if mergedUsrSystem() {
+		moddir = "/usr/lib/modules"
+	}
+
 	for _, v := range modules {
-		usrpath := "/lib/modules"
-		if mergedUsrSystem() {
-			usrpath = "/usr/lib/modules"
-		}
-		if err := w.CopyFile(path.Join(usrpath, kernelRelease, v)); err != nil {
+		if err := w.CopyFile(path.Join(moddir, kernelRelease, v)); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


obbardc:
this should close #13 and improve Ubuntu support. Thanks Andrej!

obbardc:
actually, looking into this a bit further shows two other pull requests with similar functionality:
https://github.com/go-debos/fakemachine/pull/45
https://github.com/go-debos/fakemachine/pull/36

still, all three patchsets do not support reading from `modules.builtin.bin`

andrewshadura:
To be honest, I think while my knowledge of Go is very basic, reading the file into a map seems a better choice for me than slurping it into a string, splitting it into array and do linear searches in it.

andrewshadura:
I have included xz support commits from #45 into my pull requests.

obbardc:
agreed with the above, nice work ;-)

machine.go
```diff
@@ -352,12 +353,47 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) error {
...
+	moddir := "/lib/modules"
+	if mergedUsrSystem() {
+		moddir = "/usr/lib/modules"
+	}
+
+	// build a list of built-in modules so that we don’t attempt to copy them
+	var builtinModules = make(map[string]bool)
+
+	f, err := os.Open(path.Join(moddir, kernelRelease, "modules.builtin"))
+	if err != nil {
```
obbardc: hm, if the file does not exist, we should not attempt to load `modules.builtin` and just load the modules normally ?
what about systems that use `modules.builtin.bin` ?

andrewshadura: I was considering this, but then I realised we already unconditionally copy `modules.builtin` (unless I’m misreading the code), so it is already required to be present.

Issue #21, opened 2018-05-21
mkfs.ext4 is not in the user path


elboulangero:
When run with the option `--scratchsize`, fakemachine needs to create and format a disk image. This requires `/sbin/mkfs.ext4`, which is not in the default path.

    $ fakemachine -s 1G
    fakemachine: exec: "mkfs.ext4": executable file not found in $PATH

The workaround is simple: just add `/sbin` to the path.

    $ PATH=$PATH:/sbin fakemachine -s 1G
    Running /bin/bash
    bash-4.4# 

I'm wondering if fakemachine should do that automatically, so that things work "out of the box".

----

In the code: fakemachine setups the scratch size very early

https://github.com/go-debos/fakemachine/blob/a8354cf53705cd2ec328254b8e8fdd6251af6da0/machine.go#L387-L403

Then setupscratch function is in charge of creating and formatting a disk image:

https://github.com/go-debos/fakemachine/blob/a8354cf53705cd2ec328254b8e8fdd6251af6da0/machine.go#L358-L377

elboulangero:
@sjoerd-ccu Maybe fakemachine could check some well-known paths, like `/sbin`, `/usr/sbin`, `/usr/local/sbin`? Does it sounds reasonable?

PR #50, opened 2020-04-29
machine: expand $PATH with /sbin and /usr/sbin

obbardc:
Some of the commands used (e.g mkfs.ext4) are normally located under /sbin. It's 
not typical for unprivileged users to have /sbin in their $PATH so fakemachine
should add /sbin:/usr/sbin to the $PATH.

Resolves: #21

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -434,6 +434,8 @@ func (m *Machine) cleanup() {
 func (m *Machine) startup(command string, extracontent [][2]string) (int, error) {
 	defer m.cleanup()
 
+	os.Setenv("PATH", os.Getenv("PATH") + ":/sbin:/usr/sbin")
+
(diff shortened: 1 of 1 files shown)


machine.go
```diff
@@ -379,7 +379,22 @@ func (m *Machine) setupscratch() error {
 	if err != nil {
 		return err
 	}
-	mkfs := exec.Command("mkfs.ext4", "-q", tmpfile.Name())
+
+	// search $PATH for executable; falling back to /sbin if not found
+	executable, err := exec.LookPath("mkfs.ext4")
+	if err != nil {
+		prefix := ""
+		if mergedUsrSystem() {
```
sjoerdsimons, line 387: no need to do that on merged systems; /sbin/mkfs will work just as well as /usr/sbin on those

Issue #48, opened 2020-03-02
search $PATH for excutable rather than assuming paths


obbardc:
From: @gasinvein

For Fedora support `/usr/bin/busybox` is under `/usr/sbin/busybox`. Maybe search `PATH` for it instead of checking only `/bin` and `/usr/bin`?

probably extends to other executables as well.

obbardc:
Wonder if this is related to #21 ?

PR #51, opened 2020-04-29
machine: search for busybox under $PATH

obbardc:
busybox is normally located under /bin but in fedora at least, it is located
under /sbin. So search both $PATH for the executable.

Resolves: #48

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -485,7 +485,13 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 	}
 	w.CopyFile(prefix + "/lib/x86_64-linux-gnu/libresolv.so.2")
 	w.CopyFile(prefix + "/lib/x86_64-linux-gnu/libc.so.6")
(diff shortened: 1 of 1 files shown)


machine.go
```diff
@@ -449,7 +449,16 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 	}
 	w.CopyFile(prefix + "/lib/x86_64-linux-gnu/libresolv.so.2")
 	w.CopyFile(prefix + "/lib/x86_64-linux-gnu/libc.so.6")
-	w.CopyFile(prefix + "/bin/busybox")
+
+	// search for busybox; in some distros it's located under /sbin
+	busybox, err := exec.LookPath("busybox")
+	if err != nil {
+		busybox, err = exec.LookPath(prefix + "/sbin/busybox")
```
sjoerdsimons, line 456: As this is teh second workaround for /sbin:/usr/sbin not in the PATH would it not make more sense to expand the path with those ?

PR #56, opened 2020-10-29
machine: Abstract the KVM-specific implementation details into seperate Backend interface

obbardc:
This MR defines the base for UML support for fakemachine.

End-user functionality wise no changes have been introduced with this MR; I have abstracted the machine specific implementation detail of the machine setup into its own interface and provided an implementation for KVM.

### DIFF
--- a/backend_kvm.go
+++ b/backend_kvm.go
@@ -98,6 +98,17 @@ func (b kvmBackend) KernelPath() (string, string, error) {
 	return kernelPath, moddir, nil
 }
 
+func (b kvmBackend) InitrdModules() []string {
+	return []string{"kernel/drivers/char/virtio_console.ko",
+			"kernel/drivers/virtio/virtio.ko",
+			"kernel/drivers/virtio/virtio_pci.ko",
+			"kernel/net/9p/9pnet.ko",
+			"kernel/drivers/virtio/virtio_ring.ko",
+			"kernel/fs/9p/9p.ko",
+			"kernel/net/9p/9pnet_virtio.ko",
+			"kernel/fs/fscache/fscache.ko"}
+}
+
 func (b kvmBackend) Start() (bool, error) {
 	m := b.machine
 
--- a/backend_kvm.go
+++ b/backend_kvm.go
@@ -1,0 +1,27 @@
+// +build linux
+// +build amd64
+
+package fakemachine
+
+import (
+	"os"
+)
+
+type kvmBackend struct {
+	machine *Machine
+}
+
+func newKvmBackend(m *Machine) kvmBackend {
+	return kvmBackend{machine: m}
+}
+
+func (b kvmBackend) Name() string {
+	return "kvm"
+}
+
+func (b kvmBackend) Supported() (bool, error) {
+	if _, err := os.Stat("/dev/kvm"); err != nil {
+		return false, err
+	}
+	return true, nil
+}
--- a/backend_kvm.go
+++ b/backend_kvm.go
@@ -4,10 +4,15 @@
 package fakemachine
 
 import (
+	"bytes"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
+	"path"
 	"strings"
+
+	"golang.org/x/sys/unix"
 )
 
 type kvmBackend struct {
@@ -37,10 +42,66 @@ func (b kvmBackend) QemuPath() (string, error) {
 	return exec.LookPath("qemu-system-x86_64")
 }
 
+func (b kvmBackend) hostKernelRelease() (string, error) {
+	/* First try the kernel the current system is running, but if there are no
+	 * modules for that try the latest from /lib/modules. The former works best
+	 * for systems directly running fakemachine, the latter makes sense in docker
+	 * environments */
+	var u unix.Utsname
+	if err := unix.Uname(&u); err != nil {
+		return "", err
+	}
+	release := string(u.Release[:bytes.IndexByte(u.Release[:], 0)])
+
+	if _, err := os.Stat(path.Join("/lib/modules", release)); err == nil {
+		return release, nil
+	}
+
+	files, err := ioutil.ReadDir("/lib/modules")
+	if err != nil {
+		return "", err
+	}
+
+	for i := len(files)-1; i >= 0; i-- {
+		/* Ensure the kernel name starts with a digit, in order
+		 * to filter out 'extramodules-ARCH' on ArchLinux */
+		filename := files[i].Name()
+		if filename[0] >= '0' && filename[0] <= '9' {
+			return filename, nil
+		}
+	}
+
+	return "", fmt.Errorf("No kernel found")
+}
+
+func (b kvmBackend) KernelPath() (string, string, error) {
+	kernelRelease, err := b.hostKernelRelease()
+	if err != nil {
+		return "", "", err
+	}
+
+	kernelPath := "/boot/vmlinuz-" + kernelRelease
+	if _, err := os.Stat(kernelPath); err != nil {
+		return "", "", err
+	}
+
+	moddir := "/lib/modules"
+	if mergedUsrSystem() {
+		moddir = "/usr/lib/modules"
+	}
+
+	moddir = path.Join(moddir, kernelRelease)
+	if _, err := os.Stat(moddir); err != nil {
+		return "", "", err
+	}
+
+	return kernelPath, moddir, nil
+}
+
 func (b kvmBackend) Start() (bool, error) {
 	m := b.machine
 
-	kernelPath, _, err := hostKernelPath()
+	kernelPath, _, err := b.KernelPath()
 	if err != nil {
 		return false, err
 	}
--- a/backend_kvm.go
+++ b/backend_kvm.go
@@ -143,6 +143,14 @@ func (b kvmBackend) MountParameters(mount mountPoint) (string, []string) {
 	return "9p", []string{"trans=virtio", "version=9p2000.L", "cache=loose", "msize=262144"}
 }
 
+func (b kvmBackend) InitModules() []string {
+	return []string{"virtio_pci", "virtio_console", "9pnet_virtio", "9p"}
+}
+
+func (b kvmBackend) InitStaticVolumes() []mountPoint {
+	return []mountPoint{}
+}
+
 func (b kvmBackend) Start() (bool, error) {
 	m := b.machine
 
--- a/backend_kvm.go
+++ b/backend_kvm.go
@@ -139,6 +139,10 @@ func (b kvmBackend) JobOutputTTY() string {
 	return "/dev/hvc0"
 }
 
+func (b kvmBackend) MountParameters(mount mountPoint) (string, []string) {
+	return "9p", []string{"trans=virtio", "version=9p2000.L", "cache=loose", "msize=262144"}
+}
+
 func (b kvmBackend) Start() (bool, error) {
 	m := b.machine
 
--- a/backend_kvm.go
+++ b/backend_kvm.go
@@ -109,6 +109,10 @@ func (b kvmBackend) InitrdModules() []string {
 			"kernel/fs/fscache/fscache.ko"}
 }
 
+func (b kvmBackend) NetworkdMatch() string {
+	return "e*"
+}
+
 func (b kvmBackend) Start() (bool, error) {
 	m := b.machine
 
--- a/backend_kvm.go
+++ b/backend_kvm.go
@@ -113,6 +113,19 @@ func (b kvmBackend) NetworkdMatch() string {
 	return "e*"
 }
 
+func (b kvmBackend) JobOutputTTY() string {
+	// By default we send job output to the second virtio console,
+	// reserving /dev/ttyS0 for boot messages (which we ignore)
+	// and /dev/hvc0 for possible use by systemd as a getty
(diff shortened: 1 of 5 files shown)


sjoerdsimons:
Looks like a great start; the backend interface is missing some parts though to be able to successfully implement other backends (in particular UML). 

At the very least you'll need to have an interface for helping to create the initscript; which is mostly difference in which module get and how the host filesystems are mounted (hostfs vs 9p and mount options).  Would be good if the backends could provide the parts they need rather then each generating a speciic init script for maintainability.

And then there are ofcourse the smaller/easier bits like default network interface naming and block devices

obbardc:
Pushed another revision which manages to make the backend creation & checking if supported a bit less convoluted.

backend_kvm.go
```diff
@@ -0,0 +1,136 @@
...
+func (b KvmBackend) InitScriptExtraCommands() []string {
+	return nil
+}
+
+func (b KvmBackend) NetworkdMatch() string {
+	return "e*"
+}
+
+func (b KvmBackend) CmdMountVolume(mount mountPoint) string {
+	return fmt.Sprintf("busybox mount -v -t 9p -o trans=virtio,version=9p2000.L,cache=loose,msize=262144 %s %s", mount.label, mount.machineDirectory)
```
sjoerdsimons, line 42: having the backends explicitly run busybox and set the `-v` on the mount command seems quite leaky.

obbardc: agreed, how about combining this function with `GetFstabEntry` with a new function which would return the mount type, options, etc for both mount command and fstab line?

backend_kvm.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (b KvmBackend) GetFstabEntry(mount mountPoint) string {
+	return fmt.Sprintf("%s %s 9p trans=virtio,version=9p2000.L,cache=loose,msize=262144 0 0", mount.label, mount.machineDirectory)
```
sjoerdsimons, line 46: And on top of the CmdMountVolume comments; there is obvious pattern here ;)  You could have one interface function which returns (<fstype>, <fsoptions>) and from that you can work out both the fstab entry *and* the mount command to run in busybox.

obbardc: ack, think we come to the same conclusion  ;-)

backend_kvm.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (b KvmBackend) GetImagePath(hostPath string, label string, index int) string {
+	return fmt.Sprintf("/dev/disk/by-id/virtio-%s", label)
+}
+
+func (b KvmBackend) GetJobOutputTTY(m *Machine) string {
```
sjoerdsimons, line 53: The one thing i'm wary about here is that we're exposing the full Machine object to the backend which makes the contract a bit vague. As it exposes the backend to a lot of implementation details (what properties fo the Machine should a backend take into account?). 

So it would probably be nicer to just provide a showBoot bool

backend_kvm.go
```diff
@@ -0,0 +1,136 @@
...
+	// (which we also ignore).
+	// If we are debugging, mix job output into the normal
+	// console messages instead, so we can see both.
+	if m.showBoot {
+		return "/dev/console"
+	}
+	return "/dev/hvc0"
+}
+
+func (b KvmBackend) Start(m *Machine) (int, error) {
```
sjoerdsimons, line 66: Similar to the above.. Makes me wonder whether it would make more sense to have a builder type pattern. Where the user to the backend creates one, configures it (set show boot, images and mounts points) and then queries its properties. Rather then just passing the whole machine and saying well good luck ;) (while show boot could in principle be changed behind its back between doing the GetJobOutputTTY and Start

sjoerdsimons (changes requested):
overall looks nice; I suspect you'r still missing the kernel module handling as that'll be somewhat different. As well as the fun of starting helpers (e.g. libslrip) for uml?

backend_kvm.go
obbardc: hmm i did think about start returning an array of the commands to run... but in some cases thats not enough, i.e libslirp helper for uml needs a sockerpair creating

sjoerdsimons: I think you misunderstood what i meant..

GetJobOutputTTY is basically; Please tell me what would the the tty *if* this particular option is set. With the assumption that that's the only determining factor. And ofcourse with the hidden contract that the showBoot option on the machine isn't changed.

What i'm wondering is if it wouldn't be nicer to build up a statefull backend object on which you can then query and just call start on. Which would also decouple the machine struct from the backends.. However a minimal variant could be something that results inthe follow pseudocode for the user:
```
instance := b.new_backend(machine);
tty := iinstance.GetJobOutputTTY();
imagepath := instance.GetImagePath(<image>);
instance.Start();
```
So again you'd flip things around; So where currently you ask  "imagine we'd start with <some useful options>" what would be your answer in that case and then only at Start() giving the full setup; With this change you'd start with give me an backend setup/instance for this configuration and then query things against that immutable instance.

backend_kvm.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (b KvmBackend) GetExtraVolumes() []mountPoint {
+	return nil
+}
+
+func (b KvmBackend) MountOptions(mount mountPoint) (fstype string, options []string, device string, mountpoint string) {
+	fstype = "9p"
+	options = []string{"trans=virtio", "version=9p2000.L", "cache=loose", "msize=262144"}
+	device = mount.label
+	mountpoint = mount.machineDirectory
```
sjoerdsimons, line 41: Any reason why this function need to define the mountpoint and the device/label ? I'd expect that's knowledge the machine has rather then the backend

backend_kvm.go
```diff
@@ -0,0 +1,136 @@
...
+	return
+}
+
+func (b KvmBackend) NetworkdMatch() string {
+	return "e*"
+}
+
+func (b KvmBackend) GetImagePath(img image) string {
+	return fmt.Sprintf("/dev/disk/by-id/virtio-%s", img.label)
```
sjoerdsimons, line 50: depending on the backend just the image may not be enough; virtio is nice in that it follows the labels (if formatted properly) but others may not have that and would need to depend on e.g. the index of the virtual disk.. Which comes back to the comment further down about creating a stateful instance first.

backend_kvm.go
obbardc: in the KVM 9p case, the volume is mounted by its label, but in the UML hostfs case the volume is mounted by a mount option and the device is none.
So the backend really needs to decide the fstype, device and options. But you are right; we could drop the mountpoint.

backend_kvm.go
obbardc: further on the image struct was modified to store the index of the virtual disk ;-)

sjoerdsimons: Yeah so that leaks more implementation details in common structures *and* adds the assumption that the backend will start in a way such that the virtual machine follows those indices

machine.go
```diff
@@ -337,42 +398,52 @@ func (m *Machine) kernelRelease() (string, error) {
 	return "", fmt.Errorf("No kernel found")
 }
 
-func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper) error {
-	kernelRelease, err := m.kernelRelease()
+func hostKernelPath() (string, string, error) {
```
sjoerdsimons, line 401: This and hostKernelRelease()  should a helper function in the kvm backend right? for UML they logic wouldn't apply

obbardc: my thinking here was to keep these helpers global (to fakemachine) since most other "backends" would use the host's kernel & modules, uml is an exception.

For instance (i am speculating here) a virtualbox backend would also use these helpers.

machine.go
```diff
@@ -118,6 +118,11 @@ if ! busybox test -L /bin ; then
 fi
 exec /lib/systemd/systemd
 `
+const udevRules = `
+# expose virtio-blk devices under /dev/disk/by-fakemachine-id
+KERNEL=="vd*[!0-9]", ATTRS{serial}=="?*", ENV{ID_SERIAL}="$attr{serial}", SYMLINK+="disk/by-fakemachine-id/$env{ID_SERIAL}"
+KERNEL=="vd*[0-9]", ATTRS{serial}=="?*", ENV{ID_SERIAL}="$attr{serial}", SYMLINK+="disk/by-fakemachine-id/$env{ID_SERIAL}-part%n"
```
sjoerdsimons: hmm this hardcodes that every backend will expose labels through the serial id which is rather limiting; actually as we're introducing udev rules we can also "remove" the restrictions on the label argument to be passed through..

e.g. one could imagine some backends simply would hardcode:
```
KERNEL=="vda", SYMLINK += "disk/by-fakemachine-id/badger"
KERNEL=="vda[0-9] SYMLINK+="disk/by-fakemachine-id/badger-part%n"
```

obbardc: ooh that's a good idea; right now the patch allows the backend to choose a udev match string. So I will make it so the backend can create a complete udev rule to be inserted into the fakemachine

cmd/fakemachine/main.go
```diff
@@ -79,6 +84,7 @@ func main() {
 	}
 
 	m := fakemachine.NewMachine()
+	m.SetBackend(options.Backend)
```
sjoerdsimons, line 87: Adding API to *change* the backend after creating the machine seems both unnecessary as well as risky.. Why not just have a new `NewMachineWithBackend` function to do it in one shot (and have NewMachine call that with e.g. "auto")

machine.go
sjoerdsimons: We can burn that bridge when we get to it and factor things out in a utility later on. But for now it's for sure odd that machine.go has backend specific helpers ;)

machine.go
sjoerdsimons: Right the only thing that should stay the probably stay consistant among backends is the SYMLINKS to be added. (e.g. disk/by-fakemachine-label/xxxx) for all of them and adjust the documentation for the label function a bit :p As currently it refers to exposing it as the SERIAL (which we can't guarantee), however the intention of this was to just make it possible to determine *which* block device matches what label.

machine.go
```diff
@@ -55,9 +56,25 @@ type Machine struct {
...
+}
+
 // Create a new machine object
-func NewMachine() (m *Machine) {
-	m = &Machine{memory: 2048, numcpus: runtime.NumCPU()}
+func NewMachineWithBackend(backend string) (*Machine, error) {
+	m := &Machine{backend: backend, memory: 2048, numcpus: runtime.NumCPU()}
+
+	// check the backend is supported
+	_, err := newBackend(*m)
```
sjoerdsimons, line 75: This is weird; why would you create a backend just to test it's supported.. And why does the machine have a backend string. 

Why not create the backend here; save the pointer in the Machine struct and call it from that later.

machine.go
```diff
@@ -95,8 +112,9 @@ func InMachine() (ret bool) {
 	return
 }
 
+// Checks whether at least one backend is supported
 func Supported() bool {
-	_, err := os.Stat("/dev/kvm")
+	_, err := NewMachineWithBackend("auto")
```
sjoerdsimons, line 119: This shows the awkward machine/backend relationship ... but lets refactor that later :)

machine.go
```diff
@@ -504,18 +504,10 @@ func (m Machine) startup(command string, extracontent [][2]string) (int, error)
...
-			udevRules = append(udevRules,
-				fmt.Sprintf(`KERNEL=="vd%c", SYMLINK+="disk/by-fakemachine-label/%s"`, driveLetter, img.label),
-				fmt.Sprintf(`KERNEL=="vd%c[0-9]", SYMLINK+="disk/by-fakemachine-label/%s-part%%n"`, driveLetter, img.label))
-		}
-		udevRules = append(udevRules, "")
-		w.WriteFile("/etc/udev/rules.d/61-fakemachine.rules", strings.Join(udevRules, "\n"), 0444)
-	}
+	// udev rules
+	udevRules := backend.UdevRules()
+	udevRules = append(udevRules, "")
```
sjoerdsimons, line 561: why appending a "" ? Is that  trick to ensure a final newline? If so just append it explicitly to the string after the joining :p

obbardc:
The verification of this PR was done using the comprehensive recipes from [apertis-image-recipes](https://gitlab.apertis.org/infrastructure/apertis-image-recipes) at commit hash `0006f36da005760de0cb88035d914477283ad0d3` - these recipes are good examples since they are up to date and include a range of different Debos actions.

The verification included making sure the recipes were built successfully without any missing steps or any failures reported.

## Built minimal ospacks for each arch
    debos -t architecture:amd64 ospack-minimal.yaml
    debos -t architecture:arm64 ospack-minimal.yaml
    debos -t architecture:armhf ospack-minimal.yaml

## Built SDK ospacks (requires disk-backed scratchspace)
    debos --scratchsize 10G  -t architecture:amd64 ospack-sdk.yaml


## Built minimal images for each arch ospack/arch combination
    debos image-rpi64.yaml
    debos image-uboot.yaml
    debos image-uefi.yaml


## Built SDK image (requires disk-backed scratchspace)
    debos --scratchsize 10G image-sdk.yaml


## Built sample images (which build their own ospacks, the production image tests ostree commit)
    debos sample-image-development.yaml
    debos sample-image-production.yaml
    debos sample-image-sdk.yaml

PR #57, opened 2020-11-02
Implement UML as a backend

obbardc:
For UML support, an additional backend has been created.

This backend is WIP and exists to show the interface capabilities introduced with #56

### DIFF
--- a/backend_uml.go
+++ b/backend_uml.go
@@ -1,0 +1,247 @@
+// +build linux
+// +build amd64
+
+package fakemachine
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path"
+
+	"golang.org/x/sys/unix"
+)
+
+type umlBackend struct {
+	machine *Machine
+}
+
+func newUmlBackend(m *Machine) umlBackend {
+	return umlBackend{machine: m}
+}
+
+func (b umlBackend) Name() string {
+	return "uml"
+}
+
+func (b umlBackend) Supported() (bool, error) {
+	// check the kernel exists
+	_, _, err := b.KernelPath()
+	if err != nil {
+		return false, err
+	}
+
+	// check the slirp helper exists exec.LookPath
+	if _, err := b.SlirpHelperPath(); err != nil {
+		return false, fmt.Errorf("libslirp-helper not installed")
+	}
+	return true, nil
+}
+
+func (b umlBackend) KernelPath() (string, string, error) {
+	// find the UML binary
+	kernelPath, err := exec.LookPath("linux.uml")
+	if err != nil {
+		return "", "", fmt.Errorf("user-mode-linux not installed")
+	}
+
+	// make sure the UML modules exist
+	// on non-merged usr systems the modules still reside under /usr/lib/uml
+	moddir := "/usr/lib/uml/modules"
+	if _, err := os.Stat(moddir); err != nil {
+		return "", "", fmt.Errorf("user-mode-linux modules not installed")
+	}
+
+	// find the subdirectory containing the modules for the UML release
+	modSubdirs, err := ioutil.ReadDir(moddir)
+	if err != nil {
+		return "", "", err
+	}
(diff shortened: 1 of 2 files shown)


backend_uml.go
```diff
@@ -0,0 +1,200 @@
...
+
+	b.kernelPath = "/usr/bin/linux.uml"
+	b.moduleDir = "/usr/lib/uml/modules"
+	b.slirpHelperPath = "/usr/bin/slirp-helper"
+
+	// TODO remove chris debug paths
+	base := "/home/obbardc/projects/debos/uml/"
+	b.kernelPath = base + "linux-uml/linux"
+	b.moduleDir = base + "linux-uml/modules/lib/modules"
+	b.slirpHelperPath = base + "libslirp-rs/target/debug/libslirp-helper"
```
sjoerdsimons, line 31: as things are getting close; drop these. Also for the binaries doesn't go have helpers to search in ones PATH ?

backend_uml.go
```diff
@@ -0,0 +1,200 @@
...
+	// which contains the modules from the base system
+	moduleVolume := mountPoint{hostDirectory: b.moduleDir, machineDirectory: "/lib/modules", label: "modules", static: true}
+	return []mountPoint{moduleVolume}
+}
+
+func (b umlBackend) Start() (int, error) {
+	// TODO commandWrapper forces network-wait-online & doesnt exit after network setup fail ?
+	// TODO disable lvm2-activation-generator ?
+	// TODO linux process is leftover when manually killing fakemachine. this bug also happens on KVM so should be tackled in follow-up
+	// TODO killing helper doesn't stop fakemachine
```
sjoerdsimons, line 104: can you add issues for things that are orthogonal to these PR's ? :)

obbardc: sure will ;-)

backend_uml.go
```diff
@@ -0,0 +1,200 @@
...
+	f2 := os.NewFile(uintptr(fds[1]), "")
+	if f2 == nil {
+		return -1, fmt.Errorf("socketpair: f2 is incorrect")
+	}
+	defer f2.Close()
+
+
+	// launch the slirp helper and attach to socketpair using --fd=0 argument
+	slirp_args := []string{"libslirp-helper",
+				"--fd=3",
```
sjoerdsimons, line 132: is fd=3 something to actaully hardcode or should that be dynamic ?

backend_uml.go
```diff
@@ -0,0 +1,200 @@
...
+		umlargs = append(umlargs,
+			"con1=fd:0,fd:1",
+			"con0=null",
+			"con=none")       // no other consoles
+	}
+
+	for i, img := range m.images {
+		// TODO passing image label to UML requires http://lists.infradead.org/pipermail/linux-um/2020-November/000779.html
+		umlargs = append(umlargs,
+			fmt.Sprintf("ubd%d=%s,,%s", i, img.path, img.label))
```
sjoerdsimons, line 179: with change the udev rules the label passing should be no longer needed

backend_uml.go
obbardc: i am going to go for hardcode; i can't see it changing unless the helper args change

sjoerdsimons: But this is the fd you get back from the socketpair right? Which is not something you can assume will be 3

obbardc: the fd created by the socketpair is indeed variable, which is then attached to the process using `os.ProcAttr`.
The `--fd` argument tells the helper which fd which has been attached to the process to use. It's always attached after stdout, stderr etc. Hopefully that's a bit clearer ?

sjoerdsimons: That deserved a comment at the very least and probably moving this code below the ProcAttr setup as it's very unclear where this =3 comes from and it's ok to hardcode it like that

obbardc: right i've split the code up a bit and hopefully explained it better in the comments...

sjoerdsimons (approved):

PR #58, opened 2020-11-02
Automatically choose backend based on heuristics 

obbardc:
By default, if no backend is selected, the auto backend is selected.

The auto backend prefers KVM and unavailable falls back to UML.

This MR depends on #57 (and therefore #56).

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -17,7 +17,20 @@ func newBackend(name string, m *Machine) (backend, error) {
 
 	switch name {
 	case "auto":
-		fallthrough
+		// select kvm first
+		b, kvm_err := newBackend("kvm", m)
+		if kvm_err == nil {
+			return b, nil
+		}
+
+		// falling back to uml
+		b, uml_err := newBackend("uml", m)
+		if uml_err == nil {
+			return b, nil
+		}
+
+		// no backend supported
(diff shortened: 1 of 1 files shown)


machine.go
```diff
@@ -68,14 +68,33 @@ func RegisterBackend(backend Backend) {
...
+		}
+
+		// falling back to uml
+		uml, uml_err := GetBackend("uml")
+		if uml_err == nil {
+			return uml, nil
+		}
+
+		// no backends upported
+		return nil, fmt.Errorf("%v, %v", kvm_err, uml_err)
```
sjoerdsimons, line 94: Will this result in a meaningful error for the user ?

obbardc: here's the user-facing errors when the backends are unsupported (i expect to implement the same for debos):

```
$ fakemachine
fakemachine: Backend not supported: kvm not supported: stat /dev/kvm: no such file or directory, uml not supported: slirp-helper not installed

$ fakemachine --backend kvm
fakemachine: Backend not supported: kvm not supported: stat /dev/kvm: no such file or directory

$ fakemachine --backend uml
fakemachine: Backend not supported: uml not supported: slirp-helper not installed
```

PR #61, opened 2021-04-20
machine: copy /etc/hosts in fakemachine

venkata-pyla:
sometimes when debootstrap uses custom repository to resolve its host to ip
 it requires an '/etc/hosts' file with its host to ip mapping in fakemachine
 assuming host machine '/etc/hosts' has this mapping, copy this file to fakemachine
 should work.

Signed-off-by: venkata pyla <venkata.pyla@toshiba-tsip.com>

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -549,6 +549,7 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 	w.CopyFile("/etc/passwd")
 	w.CopyFile("/etc/group")
 	w.CopyFile("/etc/nsswitch.conf")
+	w.CopyFile("/etc/hosts")
 
 	// udev rules
 	udevRules := strings.Join(backend.UdevRules(), "\n") + "\n"


sjoerdsimons:
What's the use-case behind this? Generally `/etc/hosts` should really be a relic of the past :) 

for debos; Does debootstrap copy hosts into the chroot otherwise using it for a custom repo will only half work

venkata-pyla:
Hi @sjoerdsimons , 

thank you for your valuable comments.

>What's the use-case behind this? Generally /etc/hosts should really be a relic of the past :)

It will help the users whose debootstrap mirrot URL is not get resolved and want to use host system hosts file,  in such environment it will help to setup base system (debootstrap) and then user has choice to use debos actions to set the /etc/hosts settings inside chroot.
may be we can further improvize this by passing the hosts mapping as a argument to debos (e.g. --add-hosts my.example.com:xxx.xxx.xxx.xxx), instead copying it every time to fakemachine.

>for debos; Does debootstrap copy hosts into the chroot otherwise using it for a custom repo will only half work

i think leaving this to the user is better as it copies to the final image, user any way can use debos actions to set this.

sjoerdsimons:
> Hi @sjoerdsimons ,
> 
> thank you for your valuable comments.
> 
> > What's the use-case behind this? Generally /etc/hosts should really be a relic of the past :)
> 
> It will help the users whose debootstrap mirrot URL is not get resolved and want to use host system hosts file, in such environment it will help to setup base system (debootstrap) and then user has choice to use debos actions to set the /etc/hosts settings inside chroot.
> may be we can further improvize this by passing the hosts mapping as a argument to debos (e.g. --add-hosts my.example.com:xxx.xxx.xxx.xxx), instead copying it every time to fakemachine.

If you want a local mirror just for speeding things up; the typical thing people tend to do is using apt-cacher-ng  on some local url (and set the proxy variables which debos passes along). Or just pass an ip address really.

I don't be entirely against adding the option to debos/fakemachine to add entries in the internal `/etc/hosts`  but....

> > for debos; Does debootstrap copy hosts into the chroot otherwise using it for a custom repo will only half work
> 
> i think leaving this to the user is better as it copies to the final image, user any way can use debos actions to set this.

...  the user is just building a recipe; so now to support the special setup need to change the debos recipe to copy in hosts?  That sounds like bad plan.

If we add something like this it needs to be done properly end-to-end; But really it all feels like working around a broken setup to me.

sangorrin:
Hi @sjoerdsimons 

Just to clarify the use case that @venkata-pyla has:
- He has to use a "Debian repository snapshots" server not controlled by him (not for speeding up)
- The server is accessed via https (using IP addresses fail): https://mycompany.debian.mirror.snapshots.co.jp
- The DNS server in his office sometimes works and sometimes it doesn't
- So he wants to add this /etc/hosts file to the Fakemachine's rootfs for those times when the DNS server does not resolve the URL properly
- Regarding the chroot: for some unknown reason, it works without adding it to the chroot. 

Having said that, I think we can use a workaround like the one below instead.

```
architecture: {{ $architecture }}

actions:
  - action: run
    chroot: false
    command: |
      echo "145.56.234.3  https://mycompany.debian.mirror.snapshots.co.jp mymirror" >> /etc/hosts

  - action: run
    chroot: false
    command: cat /etc/hosts

  - action: debootstrap
    suite: buster

  - action: overlays
     source: overlays/hostsfile/
```

So if there is no interest or common use, we can just use this workaround.

Thanks,
Daniel

PR #62, opened 2021-06-14
Automatically copy module dependencies

santoshmahto7:
These commits provide better support for running fakemachine on archlinux and other hosts. This pull request supersedes older https://github.com/go-debos/fakemachine/pull/41 and all the review comments from older PR are addressed.

### DIFF
--- a/backend_kvm.go
+++ b/backend_kvm.go
@@ -151,11 +151,9 @@ func (b kvmBackend) InitrdModules() []string {
 	return []string{"virtio_console",
 			"virtio",
 			"virtio_pci",
-			"9pnet",
 			"virtio_ring",
 			"9p",
-			"9pnet_virtio",
-			"fscache"}
+			"9pnet_virtio"}
 }
 
 func (b kvmBackend) UdevRules() []string {
--- a/backend_kvm.go
+++ b/backend_kvm.go
@@ -45,7 +45,7 @@ func (b kvmBackend) QemuPath() (string, error) {
 	return exec.LookPath("qemu-system-x86_64")
 }
 
-func (b kvmBackend) hostKernelRelease() (string, error) {
+func (b kvmBackend) HostKernelRelease() (string, error) {
 	/* First try the kernel the current system is running, but if there are no
 	 * modules for that try the latest from /lib/modules. The former works best
 	 * for systems directly running fakemachine, the latter makes sense in docker
@@ -124,7 +124,7 @@ func (b kvmBackend) hostKernelPath(kernelRelease string) (string, error) {
 }
 
 func (b kvmBackend) KernelPath() (string, string, error) {
-	kernelRelease, err := b.hostKernelRelease()
+	kernelRelease, err := b.HostKernelRelease()
 	if err != nil {
 		return "", "", err
 	}
--- a/backend_kvm.go
+++ b/backend_kvm.go
@@ -148,14 +148,14 @@ func (b kvmBackend) KernelPath() (string, string, error) {
 }
 
 func (b kvmBackend) InitrdModules() []string {
-	return []string{"kernel/drivers/char/virtio_console.ko",
-			"kernel/drivers/virtio/virtio.ko",
-			"kernel/drivers/virtio/virtio_pci.ko",
-			"kernel/net/9p/9pnet.ko",
-			"kernel/drivers/virtio/virtio_ring.ko",
-			"kernel/fs/9p/9p.ko",
-			"kernel/net/9p/9pnet_virtio.ko",
-			"kernel/fs/fscache/fscache.ko"}
+	return []string{"virtio_console",
+			"virtio",
+			"virtio_pci",
+			"9pnet",
+			"virtio_ring",
+			"9p",
+			"9pnet_virtio",
+			"fscache"}
 }
 
 func (b kvmBackend) UdevRules() []string {
--- a/machine.go
+++ b/machine.go
@@ -6,6 +6,7 @@ package fakemachine
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -28,6 +29,80 @@ func mergedUsrSystem() bool {
 	}
 
 	return false
+}
+
+// Parse modinfo output and return the value of module attributes
+// There may be multiple row with same fieldname so []string
+// is used to return all data.
+func getModData(modname string, fieldname string) []string {
+	out, err := exec.Command("modinfo", modname).Output()
+	if err != nil {
+		return nil
+	}
+
+	var fieldValue []string
+	scanner := bufio.NewScanner(strings.NewReader(string(out)))
+	scanner.Split(bufio.ScanLines)
+	for scanner.Scan() {
+		field := strings.Split(strings.TrimSpace(scanner.Text()), ":")
+		if strings.TrimSpace(field[0]) == fieldname {
+			fieldValue = append(fieldValue, strings.TrimSpace(field[1]))
+		}
+	}
+	return fieldValue
+}
+
+// Get full path of module
+func getModPath(modname string) string {
+	path := getModData(modname, "filename")
+	if len(path) != 0  {
+		return path[0]
+	}
+	return ""
+}
+
+// Get all dependent module
+func getModDepends(modname string) []string {
+	deplist := getModData(modname, "depends")
+	var modlist []string
+	for _, v := range deplist {
+		if  v != "" {
+			modlist = append(modlist, strings.Split(v, ",")...)
+		}
+	}
+
+	return modlist
+}
+
+func copyModules(w *writerhelper.WriterHelper, modname string, copiedModules map[string]bool) error {
+	modpath := getModPath(modname)
+	if modpath == "" {
+		return errors.New("Modules path couldn't be determined")
+	}
+
+	if modpath == "(builtin)" || copiedModules[modname] {
+		return nil
+	}
+
+	prefix := ""
+	if mergedUsrSystem() {
+		prefix = "/usr"
+	}
+
+	if err := w.CopyFile(prefix + modpath); err != nil {
+		return err
+	}
+
+	copiedModules[modname] = true;
+
+	deplist := getModDepends(modname)
+	for _, mod := range deplist {
+		if err := copyModules(w, mod, copiedModules); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // Evaluate any symbolic link, then return the path's directory. Returns an
@@ -389,13 +464,13 @@ func (m *Machine) SetEnviron(environ []string) {
 	m.Environ = environ
 }
 
+
 func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper, moddir string, modules []string) error {
 	if len(modules) == 0 {
 		return nil
 	}
 
-	modules = append(modules,
-			"modules.order",
+	modfiles := []string {"modules.order",
 			"modules.builtin",
 			"modules.dep",
 			"modules.dep.bin",
@@ -405,44 +480,18 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper, moddir strin
 			"modules.symbols",
 			"modules.symbols.bin",
 			"modules.builtin.bin",
-			"modules.devname")
+			"modules.devname"}
 
-	// build a list of built-in modules so that we don’t attempt to copy them
-	var builtinModules = make(map[string]bool)
-
-	f, err := os.Open(path.Join(moddir, "modules.builtin"))
-	if err != nil {
-		return err
-	}
-	defer f.Close()
-
-	scanner := bufio.NewScanner(f)
-	for scanner.Scan() {
-		module := scanner.Text()
-		builtinModules[module] = true
-	}
-
-	if err := scanner.Err(); err != nil {
-		return err
-	}
-
-	for _, v := range modules {
-		if builtinModules[v] {
-			continue
+	for _, v := range modfiles {
+		if err := w.CopyFile(moddir + "/" + v); err != nil {
+			return err
 		}
+	}
 
-		modpath := path.Join(moddir, v)
+	copiedModules := make(map[string]bool)
 
-		if strings.HasSuffix(modpath, ".ko") {
-			if _, err := os.Stat(modpath); err != nil {
-				modpath += ".xz"
-			}
-			if _, err := os.Stat(modpath); err != nil {
-				return err
-			}
-		}
-
-		if err := w.CopyFile(modpath); err != nil {
+	for _, modname := range modules  {
+		if err := copyModules(w, modname, copiedModules); err != nil {
 			return err
 		}
 	}
--- a/machine.go
+++ b/machine.go
@@ -34,8 +34,8 @@ func mergedUsrSystem() bool {
 // Parse modinfo output and return the value of module attributes
 // There may be multiple row with same fieldname so []string
 // is used to return all data.
-func getModData(modname string, fieldname string) []string {
-	out, err := exec.Command("modinfo", modname).Output()
+func getModData(modname string, fieldname string, kernelRelease string) []string {
+	out, err := exec.Command("modinfo", "-k", kernelRelease, modname).Output()
 	if err != nil {
 		return nil
 	}
@@ -53,8 +53,8 @@ func getModData(modname string, fieldname string) []string {
 }
 
 // Get full path of module
-func getModPath(modname string) string {
-	path := getModData(modname, "filename")
+func getModPath(modname string, kernelRelease string) string {
+	path := getModData(modname, "filename", kernelRelease)
 	if len(path) != 0  {
 		return path[0]
 	}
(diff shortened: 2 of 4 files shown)


santoshbit20071:
It would be great if someone could review this pull request, Thanks

obbardc:
Looks good to me, did you test this series on both Debian and Arch ?

backend_kvm.go
```diff
@@ -102,6 +147,7 @@ func (b kvmBackend) InitrdModules() []string {
 	return []string{"kernel/drivers/char/virtio_console.ko",
 			"kernel/drivers/virtio/virtio.ko",
 			"kernel/drivers/virtio/virtio_pci.ko",
+			"kernel/drivers/virtio/virtio_pci_modern_dev.ko",
```
obbardc, line 150: this module isnt included in debian kernel packages..

obbardc (changes requested):

backend_kvm.go
```diff
@@ -74,14 +75,58 @@ func (b kvmBackend) hostKernelRelease() (string, error) {
 	return "", fmt.Errorf("No kernel found")
 }
 
+func (b kvmBackend) hostKernelPath(kernelRelease string) (string, error) {
+	kernelDir := "/boot"
```
gasinvein, line 79: Kernel not always reside in `/boot`. For exmple, on systemd-boot using hosts, kernels are likely to end up in `/boot/efi/$machine_id/$uname_r/linux` (or `/efi`). On Fedora, `kernel` packages install the kernel at `/usr/lib/modules/$uname_r/vmlinuz`.
I guess supporting all the cases might be a lot of work, so maybe it's possible make the kernel path selectable manually via a CLI arg or an environment variable?

sjoerdsimons:
@santoshbit20071 next time could you split your patches into logical topics ; for now i'll have a first look if it's all fine then you can leave it like this

machine.go
```diff
@@ -534,17 +534,17 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
...
 	if err != nil {
 		return -1, err
 	}
-	w.CopyFileTo(busybox, prefix + "/bin/busybox")
+
+	if mergedUsrSystem() {
+		w.CopyFileTo(busybox, "/usr/bin/busybox")
+	} else {
+		w.CopyFileTo(busybox, "/bin/busybox")
+	}
```
evelikov, line 554: This is perfectly valid as-is, although it could be made simpler. Since we always have `/bin/` (as symlink in the merged case) we could use a single `w.CopyFileTo(busybox, "/bin/busybox")`

backend_kvm.go
evelikov: Fwiw the module was introduced with a while back - April 2015 with Linux v4.0. Perhaps we should mark this module as "optional" so we don't error out, when it's missing?

machine.go
obbardc: actually this patch can be dropped since it doesnt add anything; but agree you could probably write a separate cleanup patch to not use `prefix` and just dump everything under `/bin`

backend_kvm.go
evelikov: Since the existing code already assumes `/boot`, patch as-is is a good start. IIRC most distros install the kernel image in `/usr/lib/modules/$uname_r/vmlinuz` while the `/boot` or `/efi` one being a copy, during the package post-install stage.

backend_kvm.go
obbardc: lots of features are introduced in the kernel, doesn't mean that distros have to  build all of it though ;-)

This is what makes me think this series hasn't been tested properly, since this change would outright fail under debian

evelikov: Guilty - I only tried Arch, not Debian. Should we add some documentation what testing "properly" entails? Perhaps even add few CI pipelines - a selection from debos-recipes?

santoshbit20071: If modules listed in array are not available in kernel then it will be ignored for copying. This check is done by scanning modules.builtin  to check if listed driver modules are available or not.
https://github.com/go-debos/fakemachine/blob/9e6ee78d073ec35da1e0c8283c128744f1164386/machine.go#L412

backend_kvm.go
gasinvein: > install the kernel image in `/usr/lib/modules/$uname_r/vmlinuz` while the `/boot` or `/efi` one being a copy

That's accurate, at least for Fedora.

backend_kvm.go
obbardc: that is untrue - if the module isn't built then it won't be present in `builtinModules` and an error will be returned.

for a debian machine:
```
$ ls -la /lib/modules/5.10.0-6-amd64/kernel/drivers/virtio/
total 304
drwxr-xr-x  2 root root  4096 Apr 18 07:39 .
drwxr-xr-x 79 root root  4096 Apr 18 07:39 ..
-rw-r--r--  1 root root 42339 Apr  9 19:17 virtio_balloon.ko
-rw-r--r--  1 root root  7227 Apr  9 19:17 virtio_dma_buf.ko
-rw-r--r--  1 root root 22139 Apr  9 19:17 virtio_input.ko
-rw-r--r--  1 root root 28115 Apr  9 19:17 virtio.ko
-rw-r--r--  1 root root 54107 Apr  9 19:17 virtio_mem.ko
-rw-r--r--  1 root root 22115 Apr  9 19:17 virtio_mmio.ko
-rw-r--r--  1 root root 57379 Apr  9 19:17 virtio_pci.ko
-rw-r--r--  1 root root 52403 Apr  9 19:17 virtio_ring.ko


$ cat /lib/modules/5.10.0-6-amd64/modules.builtin | grep virtio | wc -l
0

```

machine.go
sjoerdsimons: This is copying into a cpio archive; not a filesystem; The suggestion from @evelikov doesn't seem safe without validating the kernel cpio follows previously unpacked symlinks in that way but even if it does it would be somewhat unexpected so i don't think that's a good suggestion

backend_kvm.go
```diff
@@ -74,14 +75,58 @@ func (b kvmBackend) hostKernelRelease() (string, error) {
...
+	if _, err := os.Stat(kernelPath); err == nil {
+		return kernelPath, nil
+	}
+
+	/* Otherwise, inspect each kernel installed, and look for the release
+	 * string straight in the binary. Not pretty, but it works. */
+	needle := kernelRelease
+	if !strings.HasSuffix(needle, " ") {
+		needle += " "
+	}
```
sjoerdsimons, line 93: the addition of a space here could use a comment/note at the very least

backend_kvm.go
```diff
@@ -74,14 +75,58 @@ func (b kvmBackend) hostKernelRelease() (string, error) {
...
+		}
+
+		kernelPath := filepath.Join(kernelDir, f.Name())
+		buf, err := ioutil.ReadFile(kernelPath)
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "Failed to read kernel:", err)
+			continue
+		}
+
+		if !bytes.Contains(buf, []byte(needle)) {
```
sjoerdsimons, line 112: that's.. interesting. Is there no existing golang library that can disect a vmlinuz file? e.g. file can:
```
$  file /boot/vmlinuz-5.10.0-3-amd64 
/boot/vmlinuz-5.10.0-3-amd64: Linux kernel x86 boot executable bzImage, version 5.10.0-3-amd64 (debian-kernel@lists.debian.org) #1 SMP Debian 5.10.17-1 (2021-02-12), RO-rootFS, swap_dev 0x6, Normal VGA
```

backend_kvm.go
evelikov: Actually having a closer look, I think we can make this much cleaner and better by parsing `modinfo` output (or using a golang library to that affect):
 - use the `filename` token - we can drop the a) hard-coded kernel module path b) the `modules.builtin` parsing and c) the compression detection/fallback code
 - use the `depends` token and error out if module is missing - we add each dependency, in the above case `virtio_pci_modern_dev` is a dependency of `virtio_pci`

Potentially for a future MR:
Add any `softdep` and `firmware` entries but do not error out, if copying fails,

backend_kvm.go
evelikov: I couldn't find any golang library for this, IIRC neither did Arnaud (original author). The Arch initrd generator `mkinitcpio` does _this exact_ search within vmlinuz and from a quick look so does `file` itself. Generally I would be weary in parsing the output of `file`. Especially since the `file` `cpio` (variant) detection has been broken for years.

From my Arch box:
```
/boot/vmlinuz-linux: Linux kernel x86 boot executable bzImage, version 5.12.6-arch1-1 (linux@archlinux) #1 SMP PREEMPT Sun, 23 May 2021 00:45:50 +0000, RO-rootFS, swap_dev 0x8, Normal VGA
/boot/vmlinuz-linux-lts: Linux kernel x86 boot executable bzImage, version 5.10.39-1-lts (linux-lts@archlinux) #1 SMP Sat, 22 May 2021 10:57:31 +0000, RO-rootFS, swap_dev 0x8, Normal VGA
```

santoshbit20071:
The review branch has been updated after addressing all review comments. The logic of copying module inside initramfs is also changed as it now uses modinfo to find the path and its dependencies
Overall previous 3 commits have been dropped and one added with these changes.
It would be great if it could be reviewed further

evelikov:
The commit message for `machine: Improve listed kernel module copying to initramfs` is a bit convoluted. Something like the following reads much better IMHO:

```
machine: handle module filename and dependencies via modinfo

Currently we do not track the proper module dependency list, but use a hard-coded list instead.
This makes our setup error prone, like seen with the recent introduction of "virtio_pci_modern_dev" an internal dependency of virtio_pci.ko.

Similarly, we also track the module path and extension. Whereby a similar internal change - moving the module to another subdir - will also break things.

Simply parse the output of modinfo, which has the added bonus of properly handling zstd supported modules as they come out.
```

backend_kvm.go
```diff
@@ -146,14 +146,14 @@ func (b kvmBackend) KernelPath() (string, string, error) {
...
-			"kernel/fs/fscache/fscache.ko"}
+	return []string{"virtio_console",
+			"virtio",
+			"virtio_pci",
+			"9pnet",
+			"virtio_ring",
+			"9p",
+			"9pnet_virtio",
+			"fscache"}
 }
```
evelikov, line 157: I suggest adding an extra commit on top, which removes the manual dependency handling above.
From a quick look - virtio, 9pnet and fscache can go.

Please mention explicitly why they are safe to remove in the commit message - X is a dependency of Y, etc.

machine.go
```diff
@@ -405,46 +448,43 @@ func (m *Machine) writerKernelModules(w *writerhelper.WriterHelper, moddir strin
...
+		if err := w.CopyFile(prefix + modpath); err != nil {
 			return err
 		}
+
+		copiedModules[modname] = true
+
+		// append dependent module of copied module
+		deplist := getModDepends(modname)
+		modules = append(modules, deplist...)
+		size += len(deplist)
```
evelikov, line 487: Thanks for fixing the dependency handling - now we properly handle the second+ level dependencies.

Not a big fan of looping over all array entries, while simultaneously adding extra ones at the end. So personally I would have made the contents of the loop a recursive standalone function. Some quick and untested example:

```
copyModule(modname...) {
    modpath := getModPath(modname)
    if modpath == "" {
        return errors.New("Modules path couldn't be determined")
    }
    if modpath == "(builtin)" || copiedModules[modname] {
        return
    }
    if err := w.CopyFile(prefix + modpath); err != nil {
        return err
    }

    copiedModules[modname] = true

    deplist := getModDepends(modname)
    for dep := range deplist {
        if err := copyModule(dep, ...); err != nil {
            return err
        }
    }
}
```

backend_kvm.go
sjoerdsimons: file know how to parse x86 kernels; See https://www.kernel.org/doc/Documentation/x86/boot.txt for the header it parses (real mode header). 

As this only happens on machine without predictable names i'm not entirely against it but ugh..

machine.go
```diff
@@ -30,6 +31,49 @@ func mergedUsrSystem() bool {
 	return false
 }
 
+// Parse modinfo output and return the value of module attributes
+// There may be multiple row with same fieldname so []string
+// is used to return all data.
+func getModData(modname string, fieldname string) []string {
+	out, err := exec.Command("modinfo", modname).Output()
```
sjoerdsimons, line 38: better to parse module.dep then he command output of modinfo here; which is fundamentally what modinfo does

evelikov: Don't think this is a good idea - the manual _explicitly_ warns against this. From `man modules.dep`

```
       Its text counterpart is located in the same directory with the name modules.dep. The
       text version is maintained only for easy of reading by humans and is in no way used
       by any kmod tool.

       These files are not intended for editing or use by any additional utilities as their
       format is subject to change in the future. You should use the modinfo(8) command to
       obtain information about modules in a future proof and compatible fashion rather than
       touching these files.
```

machine.go
santoshbit20071: done on latest push

backend_kvm.go
santoshbit20071: done on latets commit

evelikov: Did you forget `virtio` or it must be explicitly copied?

Skimming through the outstanding list - virtio and virtio_ring are lowest level modules required by anything virtio. Not sure why virtio_pci is in the list - suspect it's no (longer) needed. Perhaps something to fix with a follow-up patch (PR even).

evelikov (commented):
Regardless of the `virtio` question/suggestion the latter two patches are:
e26857e machine: Use modinfo result to handle module path and dependencies
1f759c9 Remove 9pnet and fscache from module list.
```
Reviewed-by: Emil Velikov <emil.velikov@collabora.com>
```

backend_kvm.go
santoshbit20071: virtio is not dependency of any module hardcoded above as I looked in both Arch and Debian(although it's builtin there).

So I ignored its removal suspecting it will not be copied via dependency tracing (In case copy required).

It may be that virtio(or virtio_ring) is not dependent of any modules since it is builtin in kernel and modinfo only list external modules(just assumption).  if this is the case we can remove virtio(or virtio_ring) since this will be either builtin or listed in depends. But as of now, I am not sure. Could be future patch

evelikov: It is logical for the module to not be listed as dependency, if it's already built-in.

For example - virtio and virtio_ring are built when [CONFIG_VIRTIO is set](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/virtio/Makefile?h=v5.14-rc1) while virtio_console [explicitly selects it](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/char/Kconfig?h=v5.14-rc1#n94) amongst others, yet `modinfo virtio_console` shows no dependencies.

Finishing that yak shaving can happen at later stage. This MR is good as-is.

machine.go
sjoerdsimons: Yes because command line tools are such a great binary interface. Are there no go bindings for libkmod to avoid calling out to other tools? If there aren't at the very least use it sensibly and call `modinfo -F depends <module>`  to get a direct comma seperated list

sjoerdsimons: Also you need to run modinfo -k to ensure it picks up the kernel version you're expecting it to be against rather then default to the current version

santoshmahto7: Output format of modinfo -F is not distinct across distro. e.g 
in debian
```
$modinfo -F depends watchdog
name:           watchdog
$modinfo -F depends 9p
fscache,9pnet
$modinfo -F filename 9p
/lib/modules/5.8.0-59-generic/kernel/fs/9p/9p.ko
$modinfo -F filename watchdog
name:           watchdog
(builtin)
$
```
In archbox
```
$modinfo -F depends watchdog
$modinfo -F depends 9p
fscache,9pnet
$modinfo -F filename 9p
/lib/modules/5.12.14-arch1-1/kernel/fs/9p/9p.ko.xz
$modinfo -F filename watchdog
(builtin)
$
```
But the output format of modinfo without flag is expected to be more distinct. So just to be in safe side we are not using any flag to get direct result

santoshmahto7: > Also you need to run modinfo -k to ensure it picks up the kernel version you're expecting it to be against rather then default to the current version

adding -k seems reasonable, will do it

evelikov: > Yes because command line tools are such a great binary interface.

Please don't shoot the messenger.

> Are there no go bindings for libkmod to avoid calling out to other tools?

I've found only a single project, albeit it has been unmaintained for 4 years - so not something I would recommend https://github.com/ElyKar/golang-kmod

santoshmahto7: > Also you need to run modinfo -k to ensure it picks up the kernel version you're expecting it to be against rather then default to the current version

Updated current PR with  "modinfo -k" to get the module information with the correct kernel release.

santoshmahto7: @sjoerdsimons : modinfo -k changes has been applied but ignored modinfo -F suggestion(see reply) in current updated.
Would you like to look again?

backend_kvm.go
santoshmahto7: yep we can clean at later stages

obbardc (commented):
with kernel 5.13 (from debian/experimental), debos/fakemachine now fails early with:
```
modprobe: can't load module virtio_pci_modern_dev (kernel/drivers/virtio/virtio_pci_modern_dev.ko): No such file or directory
modprobe: can't load module netfs (kernel/fs/netfs/netfs.ko): No such file or directory
```
with this PR applied, debos no longer fails early. So we really should get this in fakemachine ASAP.

PR #63, opened 2021-08-04
Use go modules

evelikov:
During go versions 1.11-1.16 the go system has gone through a notable
shift of how it handles modules/dependencies.

The original go approach was to always pull the master branch. With 1.11
things have moved to implicitly use the latest symver tag and version
1.16 prefers that dependencies are a) described in a go.mod file and b)
vendored in-tree.

Note: go 1.16 does  support externally stored dependencies, although in
that case a go.sum file must also be provided, which includes relevant
checksums.

The new approach is beneficial to both project and people building it.

The project can use, specific revision which works correctly - think
missing API, specific bugs, etc.

While the builders no longer need to track all the dependencies,
recursively, themselves and fetch them via `go get`. A simple `go build`
does everything needed.

See the following URL for more details.

https://blog.golang.org/using-go-modules
https://golang.org/ref/mod


go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/go-debos/fakemachine
+
+go 1.11
```
obbardc, line 3: how does this compare if we use a different version of go, say go 1.15 on the host (as per debian unstable) ?

go.mod
```diff
@@ -0,0 +1,10 @@
...
+
+require (
```
obbardc, line 5: it'd be nice to note how to generate these autogenerated files?

obbardc (changes requested):
couple of quick notes/questions more than anything - I assume you'll be doing the same for debos too ?

go.mod
evelikov: The 1.11 was chosen, since it's the one in Debian stable. I've only tested against Arch's 1.16. Chances are newer version might be required. Please check the official docs about the [go directive](https://golang.org/ref/mod#go-mod-file-go) - their wording if far better than any attempts of mine.

Note that the contents do not vary based on the go version.

obbardc: yup just read up on the config key; thanks for the link!

go.mod
evelikov: There's a ton of information to unpack here, I'll try to keep it brief:

Most/all notes in the go.sum and go.mod files get discarded on updates - so an inline comment is no-go.
Personally I created the go.mod manually - the dependencies were known from the existing PKGBUILD.

Then created the go.sum via
```
export GOFLAGS="-buildmode=pie -trimpath -mod=mod -modcacherw"
go build ./cmd/fakemachine/main.go # this step downloads the modules and generates go.sum
```

NOTE: With 1.16 (or was it 1.15) the default is `-mod=readonly` - which is recommended for 99% of the uses. Yet since module info is read-only go refuses (hard fails) when updating the go.sum file - hence `-mod=mod`.

In general, the linked documentation suggests using:
```
go mod init github.com/go-debos/fakemachine
go tidy
```
I've avoided that due to two reasons - a) uses `go 1.16` which is not available on Debian and b) uses the latest gocpio symver gocpio tag `v1.0.1`*

* That gocpio is broken - doesn't build. And the latest tag `1.1.0` is not symver (missing leading v) - [fix requested](https://github.com/surma/gocpio/issues/6). Go also refuses the to use the tag and creates it's own pseudo tag/branch. 

All that said - let me know if you want any of this in the commit message,

evelikov:
Re-running `go mod tidy` high-lights that I've missed the test modules. Left the commit separate to high-light the difference.
Will squash appropriately alongside the extra information request - once know what/how much people want to me add.

evelikov:
@obbardc does the above sound reasonable?

evelikov:
If I understand the go modules design correctly the concepts of auto-updating or even semi-automated the modules go against it. You might want to go through the documentation, if you haven't already.

That said - the updated commit has explicit steps, following the official docs instead of my half-hacked approach.
It should be good starting point and anyone curious for more, should really read the docs.

evelikov:
@obbardc humble ping?

evelikov:
Humble monthly ping

go.mod
```diff
@@ -0,0 +1,11 @@
...
+	github.com/docker/go-units v0.4.0
+	github.com/jessevdk/go-flags v1.5.0
+	github.com/stretchr/testify v1.7.0
+	github.com/surma/gocpio v1.0.2-0.20130630172628-c01aebd00e02
```
obbardc, line 9: you also need to add another step in the commit MR to run `go mod tidy` after editing the tag :-)

evelikov: it's already there:
```
1. go mod init github.com/go-debos/fakemachine
2. go mod tidy
3. edit go.mod as outlined below
4. re-run, go mod tidy
```

obbardc:
@evelikov I spent some time yesterday having a look at go modules and exploring the various options.

I think to solve the gocpio issue in the short term, rather than tagging our own release in our own fork (which will need packaging changes, at least for debian) we should do the following to create the go.mod:

```
go mod init github.com/go-debos/fakemachine
go mod tidy
go get github.com/surma/gocpio@c01aebd
```

Can you update the MR with the above, please ?

evelikov-work:
Ack, updated. Also added an explicit note that I've tested it with 1.16 and 1.18 - 1.11 is too old to test without spending serious time.

obbardc (approved):

PR #67, opened 2021-09-09
uml: lookup path

gportay:
Hello,

Most distros ship the binaries at different path locations.

This PR intends to make the UML backend less distro-specific by lookup the binaries in a hardcoded list of known locations.

Better idea?

Best Regards,
Gaël

### DIFF
--- a/backend_uml.go
+++ b/backend_uml.go
@@ -4,9 +4,9 @@
 package fakemachine
 
 import (
+	"strings"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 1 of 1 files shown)


evelikov:
The MR changes things in backward incompatible manner - `/usr/lib/uml/modules` magically becomes `/usr/lib/modules/uml`

gportay:
> The MR changes things in backward incompatible manner - `/usr/lib/uml/modules` magically becomes `/usr/lib/modules/uml`

Oops :facepalm:

Fixed

evelikov:
UML is very low on my priority list, since there's no official support in Arch. Will get to it eventually though.

obbardc:
Moving this forward as part of #112 so closing in favour of that PR.

PR #68, opened 2021-09-10
backend: remove InitrdModules() in favour of InitModules()

evelikov:
With recently the former was reworked from list of modules and their
respective dependencies (all of which including relative path and
extension) to a modules that we use.

The only difference between the two is - virtio and virtio_ring. Both of
those are low level dependencies of the other virtio modules.

So just remove the extra function/array and consistently use
InitModules() across the board.

As follow-up on @santoshmahto7 's work in https://github.com/go-debos/fakemachine/pull/62

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -72,9 +72,6 @@ type backend interface {
 	// The path to the modules
 	ModulePath() (moddir string, err error)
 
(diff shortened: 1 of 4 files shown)


obbardc (approved):
LGTM!

PR #70, opened 2021-09-28
Replace Arch specific kernel copy hack with more generic solution

evelikov:
This MR reverts the Arch specific commit, splits `KernelPath()` into `KernelPath()` & `ModulePath()`, reusing the last one to implement a shorter and more common solution.

In particular: Many platforms store the kernel under `/usr/lib/modules/$uname_r/vmlinuz` - perhaps because systemd expects it, to enable hibernate.

As added bonus - there is no need for a initramdisk generator, which tends to be responsible for copying the kernel in `/boot`.

Tl:Dr: Shorter, more generic code and fewer dependencies \o/

Big shout out to @gasinvein for the original suggestion.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -66,8 +66,11 @@ type backend interface {
 	// Get kernel release version
 	KernelRelease() (string, error)
 
-	// The path to the kernel and modules
-	KernelPath() (kernelPath string, moddir string, err error)
+	// The path to the kernel
+	KernelPath() (kernelPath string, err error)
+
+	// The path to the modules
+	ModulePath() (moddir string, err error)
 
 	// A list of modules to include in the initrd
 	InitrdModules() []string
--- a/backend_qemu.go
+++ b/backend_qemu.go
@@ -71,15 +71,24 @@ func (b qemuBackend) KernelRelease() (string, error) {
 	return "", fmt.Errorf("No kernel found")
 }
 
-func (b qemuBackend) KernelPath() (string, string, error) {
+func (b qemuBackend) KernelPath() (string, error) {
 	kernelRelease, err := b.KernelRelease()
 	if err != nil {
-		return "", "", err
+		return "", err
 	}
 
 	kernelPath := "/boot/vmlinuz-" + kernelRelease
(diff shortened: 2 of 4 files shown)


obbardc (approved):
untested but LGTM!

evelikov:
Thanks for the prompt reply. If you can tell me what is needed to get things merged, that would be appreciated.

Even trivial MRs like https://github.com/go-debos/fakemachine/pull/68 remain open.

gasinvein:
For me on Fedora 34 it instantly falls back to looking for kernels `/boot`.
Using `strace` I can see it tries to access `/usr/lib/modules/5.13.19-200.fc34.x86_64/5.13.19-200.fc34.x86_64/vmlinuz` - which doesn't exist.

backend_kvm.go
```diff
@@ -77,61 +76,38 @@ func (b kvmBackend) KernelRelease() (string, error) {
...
-		if !bytes.Contains(buf, []byte(needle)) {
-			continue
+	/* First we look within the modules directory, as supported by
+	 * various distributions - Arch, Debian, Fedora...
+	 *
+	 * ... perhaps because systemd requires it to allow hibernation
+	 * https://github.com/systemd/systemd/commit/edda44605f06a41fb86b7ab8128dcf99161d2344
+	 */
+	if moddir, err := b.ModulePath(); err == nil {
+		kernelPath := path.Join(moddir, kernelRelease, "vmlinuz")
```
gasinvein, line 92: Probably `kernelRelease` is redundant here? `moddir` already include kernel release subdir in its path.

evelikov: D'oh sorry about that - fixed with v2.

Note to self: don't submit 3 MRs at the same time - you're bound to miss something.

gasinvein:
Confirming this works on Fedora 34.

obbardc:
@evelikov Can you rebase this on latest main please ?

evelikov:
Done, needed a double-take since master was silently renamed to main.
Don't mind which name is used, yet a PSA would have been useful.

obbardc (approved):

PR #71, opened 2021-09-28
machine: copy only selected modules.* files

evelikov:
Busybox uses only the modules.{builtin,dep,alias,symbols}.

Any other files, such as the .bin ones are kmod specific, so we don't
need to bother with them.


obbardc (approved):
Untested but LGTM!

PR #73, opened 2022-01-07
Check mountpoints exist on the host early

obbardc:
If a mountpoint doesn't exist on the host, the backend
will fail to mount it and results in an ugly error
from the backend:

    $ fakemachine -v /null
    qemu-system-x86_64: -virtfs local,mount_tag=virtfs-5,path=/null,security_model=none: cannot initialize fsdev 'virtfs-5': failed to open '/null': No such file or directory
    fakemachine: error starting kvm backend: <nil>

Check if a mountpoint exists before attempting to start
the backend.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -13,12 +13,13 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
 	"text/template"
 
-	"github.com/go-debos/fakemachine/cpio"
+	writerhelper "github.com/go-debos/fakemachine/cpio"
 )
 
 func mergedUsrSystem() bool {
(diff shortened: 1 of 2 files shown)


machine.go
```diff
@@ -535,6 +535,12 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 
 	os.Setenv("PATH", os.Getenv("PATH") + ":/sbin:/usr/sbin")
 
+	for _, v := range m.mounts {
+		if _, err := os.Stat(v.hostDirectory); err != nil {
```
obbardc, line 539: some thoughts:
- does this stop if the volume `hostDirectory`
- does `machineDirectory` need to exist ?
- check if `machineDirectory` contains spaces
- check if `label` contains spaces
- add a unit test to check this behaviour

sjoerdsimons:
Also rebase on latest and add some tests please for the various situations ;)

PR #74, opened 2022-01-07
Rework panic calls into bubbled-up errors

obbardc:
Rather than panicking, bubble up errors such that
the execution may recover correctly and the error
may be shown to the downstream user.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -150,12 +150,8 @@ type Machine struct {
 }
 
 // Create a new machine object with the auto backend
-func NewMachine() *Machine {
-	m, err := NewMachineWithBackend("auto")
-	if err != nil {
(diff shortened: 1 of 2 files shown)


machine.go
```diff
@@ -150,12 +150,8 @@ type Machine struct {
 }
 
 // Create a new machine object with the auto backend
-func NewMachine() *Machine {
-	m, err := NewMachineWithBackend("auto")
-	if err != nil {
-		panic(err)
-	}
-	return m
+func NewMachine() (*Machine, error) {
```
obbardc, line 153: this API isn't used by debos or fakemachine, so it is OK to modify
also, this should be used for the `CreateMachine` function in #76

PR #78, opened 2022-04-06
Add lint workflow

sjoerdsimons:


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,20 @@
...
+permissions:
+  contents: read
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v2
+      - uses: actions/checkout@v2
+      - name: golangci-lint
```
obbardc, line 19: just curious if action this actually calls `gofmt` too ?
According to https://golangci-lint.run/usage/linters/ only minimal linters are enabled by default

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,20 @@
+name: golangci-lint
```
obbardc, line 1: Would it be better to split the CI workflow into a separate commit on top of fixing the lint errors ?

obbardc (changes requested):

.github/workflows/golangci-lint.yml
evelikov: IMHO it would be great to at least split the cosmetic/noop changes (whitespace, semi-colon) from the functional ones.
Using a `WriteDirectory` loop in `startup` would also be nice to have.

Regardless of the above, as a whole I'm really glad to see this PR. Thank you

sjoerdsimons: None of this is done yet;  I created the PR to test whether the workflow worked as expected ; switched to draft status to make that more clear

.github/workflows/golangci-lint.yml
sjoerdsimons: annoyingly it doesn't ; need to enable some more indeed

cpio/writerhelper.go
```diff
@@ -89,13 +143,24 @@ func (w *WriterHelper) WriteSymlink(target, link string, perm os.FileMode) {
...
+	if err != nil {
+		return err
+	}
+
+	_, err = w.Write(content)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
evelikov-work, line 156: Another style permutation - as seen in `WriteFileRaw` just above.

```
_, err = w.Write(content)
return err
```

evelikov-work:
Above all - thanks for fixing my style bugs.

I've put forward couple of style preference suggestions ... well questions really. Neither has to be fixed here or now, but it's great to have your view on the topic.

cpio/writerhelper.go
```diff
@@ -42,30 +53,55 @@ func (w *WriterHelper) ensureBaseDirectory(directory string) {
 			continue
 		}
 
-		w.WriteDirectory(collector, 0755)
+		err := w.WriteDirectory(collector, 0755)
```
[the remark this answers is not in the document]
sjoerdsimons: There isn't a general go guideline about when to use that variant; the most concrete i've seen is to not use it if the practical `foo` is long; Which is likely also why e.g. gofmt -s doesn't do those trivial rewrites.

I don't really see the point of deciding on one vs. the other or defining rules about when which form should be used.

.github/workflows/golangci-lint.yml
sjoerdsimons: fwiw gofmt lint is enabled now

PR #79, opened 2022-04-08
Run testsuite for each supported backend

obbardc:
Currently the tests only run for one backend.
Wrap the tests into a test suite and run the tests for each supported backend.

### DIFF
--- a/machine_test.go
+++ b/machine_test.go
@@ -3,51 +3,58 @@ package fakemachine
 import (
 	"bufio"
 	"flag"
-	"github.com/stretchr/testify/assert"
 	"io"
 	"os"
 	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


machine_test.go
```diff
@@ -121,55 +130,68 @@ fi
...
 	}
 }
+
+func TestFakemachineTestSuite(t *testing.T) {
+	// TODO run some tests outside of fakemachine first - finding modules etc?
+	// TODO test auto backend resolution
+	// TODO try to speed up the test by bringing up just one fakemachine per backend?
+
+	// test all backends
+	suite.Run(t, &FakemachineTestSuite{backend: "auto"})
```
obbardc, line 194: maybe have all commands which run inside a backend called `FakemachineBackendTestSuite` or similar to allow tests which don't need a backend to be ran ?

PR #80, opened 2022-04-09
Add qemu backend run actual fakemachime tests

sjoerdsimons:


obbardc:
Looks good to me !!
Hopefully the container framework means other interested parties (e.g. arch/fedora/ubuntu etc) may also add their own test containers ;-)

sjoerdsimons:
> Happy to add an Arch container - is there any documentation/references how to do that?

You send an MR with a  Dockerfile  && update to the github workflow for e.g. fakermachine-arch to  https://github.com/go-debos/test-containers ; That will build a cntainer to test against.  

Then in the MR against fakemachine add a variant with that container in the matrix and standard testing should run.

evelikov:
Thanks. Arch should be handled with:
- https://github.com/go-debos/test-containers/pull/1
- https://github.com/go-debos/fakemachine/pull/82

PR #72, opened 2021-09-28
Support zst compressed modules

evelikov:
In a gist, with linux 5.13 the kernel can produce `.zst` compressed modules, which currently fail with busybox. While there is some WIP to change that - many other init ramdisk generators extract said files (dracut, mkinitcpio).

As such we need to:
 - learn how to generate the modules.dep file - we do it manually, in the future we might delegate to `depmod` (the tool used by distros)
 - add basic decompressor for said formats
 - wire the decompressor during modules.dep generation (must strip the suffix) and module copy

This is the original work by Ed (@eds-collabora) which I've split from a single large commit, tweaked the commit message and applied few minor changes - as noted in each patch.

https://github.com/go-debos/fakemachine/issues/69

PR #82, opened 2022-04-12
Add Arch variant to the test matrix

evelikov:
Depends on https://github.com/go-debos/test-containers/pull/1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         variant:
+          - fakemachine-arch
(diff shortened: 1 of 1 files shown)


evelikov:
Converting to draft, since Arch uses zst compressed modules. Aka https://github.com/go-debos/fakemachine/pull/72 need to be merged first.

evelikov:
With the zst support (and fixed vmlinuz handling) the CI should just work with Arch.

obbardc:
@evelikov Obviously github doesn't have a rebase button (unless I merge it too, but i wanna see a successful test run), so if you could rebase on main that would be great :-)

PR #83, opened 2022-04-12
backend: Add qemu to the list of backend names

evelikov:
Otherwise the --help screen will not print it, even though we'll accept
the value at runtime.

@sjoerdsimons was it omitted deliberately?

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -9,7 +9,7 @@ import(
 
 // A list of backends which are implemented
 func BackendNames() []string {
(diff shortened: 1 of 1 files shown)


obbardc:
Good catch, it is repeated in the next function so probably just missed out: https://github.com/go-debos/fakemachine/blob/main/backend.go#L15

So perhaps it would be good to improve this in future with a table so it is defined in one place ?

PR #86, opened 2022-04-16
Automatically choose backend backed on priority

obbardc:
When new backends are added to Fakemachine there are multiple
places in the code where the new backend must be added. Let's
reduce the ability for human errors in these cases and have
just one mapping of backend vs. the priority in which it should
be chosen in the automatic backend selection algorithm.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -7,43 +7,59 @@ import (
 	"fmt"
 )
 
-// A list of backends which are implemented
-func BackendNames() []string {
-	return []string{"auto", "kvm", "uml", "qemu"}
+// List of backends in order of their priority in the "auto" algorithm
+func implementedBackends(m *Machine) []backend {
+	return []backend{
+		newKvmBackend(m),
+		newUmlBackend(m),
+		newQemuBackend(m),
+	}
 }
 
+/* A list of backends which are implemented - sorted in order in which the
+ * "auto" backend chooses them.
+ */
+func BackendNames() []string {
+	names := []string{"auto"}
+
+	for _, backend := range implementedBackends(nil) {
+		names = append(names, backend.Name())
+	}
+
+	return names
+}
+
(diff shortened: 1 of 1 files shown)


backend.go
```diff
@@ -5,45 +5,78 @@ package fakemachine
...
+func implementedBackends(m *Machine) map[int]backend {
+	return map[int]backend{
+		10: newKvmBackend(m),
+		20: newUmlBackend(m),
+		30: newQemuBackend(m),
+	}
+}
+
+/* maps are iterated randomly in Go - calculate and store the sorted backend
+ * order in a seperate array
```
evelikov, line 21: Nit: s/seperate/separate/
Another go maps trivia - they cannot be const :confused: 

In case it matters: I like the algorithmic approach but I wonder if it's worth it. Are we having for (m)any new backends in the future?

obbardc: Fixed the spelling mistake.
I added this approach due to when the qemu backend was added it was missed out from the BackendNames() function - there are no real plans to add other backends but I still think this is useful to generally improve the code as the priorities may also change in future?

.... other than my idea/hack implementation of a virtualbox backend i don't think there are many other options? ;-)

backend.go
```diff
@@ -5,45 +5,78 @@ package fakemachine
...
 )
 
-// A list of backends which are implemented
+// Main list of backends vs. their order in the "auto" algoritmn.
+func implementedBackends(m *Machine) map[int]backend {
+	return map[int]backend{
+		10: newKvmBackend(m),
+		20: newUmlBackend(m),
+		30: newQemuBackend(m),
+	}
```
sjoerdsimons, line 17: any reason for this being a map. Judging by the remainder of the code a simple array in priority order seems more suilted?

obbardc: i will rework it to an array ;-)

PR #87, opened 2022-04-16
Run CI tests on user-mode-linux backend on Debian

obbardc:
Currently we only run the tests on the qemu backend, which
misses all of the user-mode-linux backend functionality;
specifically fakemachine setup/teardown steps. We need to be
able to reliably test all backends in CI to be able to merge
any related changes with any confidence.

For now, only testing on the Debian base containers is
supported. More container baselines may be added as follow-up
pull requests.

Requires https://github.com/go-debos/test-containers/pull/11

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,16 +20,31 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        variant:
-          - fakemachine-arch
-          - fakemachine-bookworm
-          - fakemachine-bullseye
+        # Currently nested virtualisation (hence kvm) is not supported on GitHub
+        # actions; but the qemu backend is enough to test Fakemachine
+        # functionality without hardware acceleration since the majority of code
+        # is shared between the qemu and kvm backends.
+        # See https://github.com/actions/runner-images/issues/183
+        # 
+        # For Arch Linux uml is not yet supported, so only test under qemu there.
+        os: [bullseye, bookworm]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -37,8 +37,19 @@ jobs:
...
+    - name: Run unit tests (auto backend)
+      run: go test -v ./... | tee -a test.out
+
+    # kvm backend isn't supported in GitHub actions
+    # see: https://github.com/actions/runner-images/issues/183
+    #- name: Run unit tests (kvm backend)
+    #  run: go test -v ./... --backend=kvm | tee -a test.out
+
+    - name: Run unit tests (uml backend)
+      run: go test -v ./... --backend=uml | tee -a test.out
```
obbardc, line 49: Potentially this would be better in the matrix which would allow for some parallelism ?

.github/workflows/ci.yml
```diff
@@ -20,25 +20,33 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        variant:
-          - fakemachine-arch
-          - fakemachine-bookworm
-          - fakemachine-bullseye
+        os: [bullseye, bookworm]
+        backend: [qemu, uml]
```
evelikov, line 24: If there's a way, I would add a comment why kvm isn't in the list. Similarly for the Arch below (as mentioned offline UML is not officially supported in Arch just yet)

PR #88, opened 2022-04-17
Update go modules

obbardc:
Since dependabot doesn't update pseudo-version dependencies, dependabot
currently only handles updating dependencies which are tagged as semver,
manually update the go modules.

See: https://github.com/dependabot/dependabot-core/issues/3017


go.sum
```diff
@@ -13,11 +13,15 @@ github.com/stretchr/testify v1.7.1 h1:5TQK59W5E3v0r2duFAb7P95B6hEeOyEnHRa8MjYSMT
 github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/surma/gocpio v1.0.2-0.20130630172628-c01aebd00e02 h1:xhjtJ97hccFQZf4wmBcsVGUmCAmtszrrQ8C5yoK4QmQ=
 github.com/surma/gocpio v1.0.2-0.20130630172628-c01aebd00e02/go.mod h1:zaLNaN+EDnfSnNdWPJJf9OZxWF817w5dt8JNzF9LCVI=
```
evelikov, line 15: Seems off: The `go mod tidy` command should remove the old references - the gocpio above and x/sys below.

PR #105, opened 2022-09-26
backend_qemu: Avoid file ID collisions when multiple mounts share the same host path

evelikov:
Currently we get a useful warning like:

9p: Multiple devices detected in same VirtFS export, which might lead to file ID collisions and severe misbehaviours on guest!
You should either use a separate export for each device shared from host or use virtfs option 'multidevs=remap'!

So let's add the option to avoid the issue. Requires QEMU >= 4.2.0

[Emil Velikov]
 - Rebase, tweak commit message

### DIFF
--- a/backend_qemu.go
+++ b/backend_qemu.go
@@ -216,7 +216,7 @@ func (b qemuBackend) StartQemu(kvm bool) (bool, error) {
 
 	for _, point := range m.mounts {
 		qemuargs = append(qemuargs, "-virtfs",
-			fmt.Sprintf("local,mount_tag=%s,path=%s,security_model=none",
+			fmt.Sprintf("local,mount_tag=%s,path=%s,security_model=none,multidevs=remap",
 				point.label, point.hostDirectory))
 	}
 


obbardc (changes requested):
With the commit message changed to `backend_qemu: ...`, look good to me!

obbardc (approved):

obbardc (changes requested):
Commit message is a bit lacking;

```
backend_qemu: Avoid file ID collisions when multiple mounts share the same host path

When multiple mounts share the same host path, qemu warns that this situation may lead to ID collisions:

    9p: Multiple devices detected in same VirtFS export, which might lead to file
    ID collisions and severe misbehaviours on guest!
    You should either use a separate export for each device shared from host or
    use virtfs option 'multidevs=remap'!

Setting the multidevs=remap argument for the block device allows sharing more than one device per export by remapping inodes from host to guest appropriately. This change requires QEMU >= 4.2.0.
```

evelikov:
All done - sorry for the delay.

PR #106, opened 2022-09-26
Simplify module handling

evelikov:
Couple of small changes making the module.dep generation code a little cleaner and simpler. Namely:
 - stop passing extension/suffix array as argument - it's constant data
 - remove unneeded sorting of modules.dep contents

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -71,7 +71,14 @@ func getModDepends(modname string, kernelRelease string) []string {
 	return modlist
 }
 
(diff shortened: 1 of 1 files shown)


obbardc (approved):
looks good in theory; seems like no real functionality changes?

evelikov:
The contents of the modules.dep may be unsorted with this MR.
In my testing that's not an issue and kmod doesn't bother sorting either.

PR #107, opened 2022-09-26
Add diskswap

evelikov:
This PR adds support for disk backed swap. While that in itself is self-explanatory there are couple of notable points - what brought us here and some implementation details:

 - We've been using debos with dracut, which in itself uses `/var/tmp` instead of `/tmp` for its temporary folder. That in itself is not an issue, until you consider that `/var` may be size constrained for various reasons
 
 - As we can see in the `SetSwap()` comment - the `path` argument is unused, although the API is written to align with `SetScratch()`. We can keep thing as-is, drop `path` in the new API or in both. In the last case, we're breaking the public API so might want to bump the major.
 
There's no preference on my end we the implementation details point.

### DIFF
--- a/cmd/fakemachine/main.go
+++ b/cmd/fakemachine/main.go
@@ -17,6 +17,7 @@ type Options struct {
 	Memory      int               `short:"m" long:"memory" description:"Amount of memory for the fakemachine in megabytes"`
 	CPUs        int               `short:"c" long:"cpus" description:"Number of CPUs for the fakemachine"`
 	ScratchSize string            `short:"s" long:"scratchsize" description:"On-disk scratch space size (with a unit suffix, e.g. 4G); if unset, memory backed scratch space is used"`
+	SwapSize    string            `long:"swapsize" description:"On-disk swap space size (with a unit suffix, e.g. 2G); if unset, swap size will be zero"`
 	ShowBoot    bool              `long:"show-boot" description:"Show boot/console messages from the fakemachine"`
 }
 
@@ -164,6 +165,15 @@ func main() {
 			os.Exit(1)
 		}
 		m.SetScratch(size, "")
+	}
+
+	if options.SwapSize != "" {
+		size, err := units.FromHumanSize(options.SwapSize)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "fakemachine: Couldn't parse swap size: %v\n", err)
+			os.Exit(1)
+		}
+		m.SetSwap(size, "")
 	}
 
 	if options.Memory > 0 {
--- a/machine.go
+++ b/machine.go
@@ -169,6 +169,11 @@ type Machine struct {
 	scratchfile string
 	scratchdev  string
 	initrdpath  string
+
+	swapsize int64
+	swappath string
+	swapfile string
+	swapdev  string
 }
 
 // Create a new machine object with the auto backend
@@ -458,6 +463,20 @@ func (m *Machine) SetScratch(scratchsize int64, path string) {
 	}
 }
 
+// SetSwap sets the size and location of on-disk swap space to allocate
+// for fakemachine VM. If not set, swap space will not be allocated. If
+// Path is "" then the working directory is used as a default storage location
+// Since this function has single caller which uses empty path, path argument
+// is not required but still added to keep in symmnetry with SetScratch.
+func (m *Machine) SetSwap(swapsize int64, path string) {
+	m.swapsize = swapsize
+	if path == "" {
+		m.swappath, _ = os.Getwd()
+	} else {
+		m.swappath = path
+	}
+}
+
 func (m Machine) generateFstab(w *writerhelper.WriterHelper, backend backend) error {
 	fstab := []string{"# Generated fstab file by fakemachine"}
 
@@ -466,6 +485,11 @@ func (m Machine) generateFstab(w *writerhelper.WriterHelper, backend backend) er
 	} else {
 		fstab = append(fstab, fmt.Sprintf("%s /scratch ext4 defaults,relatime 0 0",
 			m.scratchdev))
+	}
+
+	if m.swapfile != "" {
+		fstab = append(fstab, fmt.Sprintf("%s none swap defaults 0 0",
+			m.swapdev))
 	}
 
 	for _, point := range m.mounts {
@@ -563,12 +587,36 @@ func (m *Machine) setupscratch() error {
 	return err
 }
 
+func (m *Machine) setupswap() error {
+	if m.swapsize == 0 {
+		return nil
+	}
+
+	tmpfile, err := ioutil.TempFile(m.swappath, "fake-swap.img.")
+	if err != nil {
+		return err
+	}
+	m.swapfile = tmpfile.Name()
+
+	m.swapdev, err = m.CreateImageWithLabel(tmpfile.Name(), m.swapsize, "fake-swap")
+	if err != nil {
+		return err
+	}
+	mkswap := exec.Command("mkswap", tmpfile.Name())
+	return mkswap.Run()
+}
+
 func (m *Machine) cleanup() {
 	if m.scratchfile != "" {
 		os.Remove(m.scratchfile)
 	}
 
+	if m.swapfile != "" {
+		os.Remove(m.swapfile)
+	}
+
 	m.scratchfile = ""
+	m.swapfile = ""
 }
 
 // Start the machine running the given command and adding the extra content to
@@ -585,8 +633,11 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 	m.AddVolumeAt(tmpdir, "/run/fakemachine")
 	defer os.RemoveAll(tmpdir)
 
-	err = m.setupscratch()
-	if err != nil {
+	if err := m.setupscratch(); err != nil {
+		return -1, err
+	}
+
+	if err := m.setupswap(); err != nil {
 		return -1, err
 	}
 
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,2 +1,2 @@
-fakemachine
+/fakemachine
```
obbardc, line 1: can you please split into another PR

cmd/fakemachine/main.go
```diff
@@ -17,6 +17,7 @@ type Options struct {
 	Memory      int               `short:"m" long:"memory" description:"Amount of memory for the fakemachine in megabytes"`
 	CPUs        int               `short:"c" long:"cpus" description:"Number of CPUs for the fakemachine"`
 	ScratchSize string            `short:"s" long:"scratchsize" description:"On-disk scratch space size (with a unit suffix, e.g. 4G); if unset, memory backed scratch space is used"`
+	SwapSize    string            `long:"swapsize" description:"On-disk swap space size (with a unit suffix, e.g. 2G); if unset, swap size will be zero"`
```
obbardc, line 20: playing devils advocate here; but what's a typical usecase of this new parameter?

obbardc (changes requested):

cmd/fakemachine/main.go
evelikov: I've tried to mentioned that with the dracut example above, but forgot the final sentence :facepalm: The complete version is:

We've redirected dracut to use /tmp (instead of it's usual /var/tmp, since this one is size constrained) but this is causing random OOM issues. That's due to two factors to:
 - outstanding artefacts in /tmp
 - un-compressing and re-compressing kernel modules and firmware - latter in itself easily pushing few GBs
 
That's my understanding at least - as you can see I'm not the one who debugged or wrote the patch. @santoshmahto7 feel free to correct me or provide any details I've missed.

obbardc: > We've redirected dracut to use /tmp (instead of it's usual /var/tmp, since this one is size constrained) but this is causing random OOM issues. That's due to two factors:
> 
> * outstanding artefacts in /tmp
> * un-compressing and re-compressing kernel modules and firmware - latter in itself easily pushing few GBs
> 
> That's my understanding at least - as you can see I'm not the one who debugged or wrote the patch. @santoshmahto7 feel free to correct me or provide any details I've missed.

Right - by default `/scratch` (and hence the rootfs) is by default backed by RAM, which is just shy of 2G by default. In these cases, we usually use `debos --scratchsize` to back the scratch area on disk.
@santoshmahto7 can you confirm whether using a scratchsize would suit your approach, or whether the swap space is needed for some other reason ?

santoshmahto7: For our use case, we are already using disk backed scratch (with size=36GB).   
The problem is fakemachine RAM. By default, it is 2G (can be increased wth -m ) but when fakemachine itself is running in low - end system then there is no way to increase RAM beyond host limit, so we need to back it with swap space.   

So this patch provides the way to set swap space (disk backed) thus setting up virtual memory for fakemachine, otherwise every process running in fakemachine uses pure ram - even opening big size file will fail.

santoshmahto7: yeah this is swap space support for fakemachine.

evelikov: Happy to tweak the commit message in any shape and form, please provide concrete suggestions.
Alternatively @obbardc you should have commit access to the branch if you want to handle it yourself.

machine.go
```diff
@@ -600,8 +648,12 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
...
 	defer os.RemoveAll(tmpdir)
 
-	err = m.setupscratch()
-	if err != nil {
+	if err := m.setupscratch(); err != nil {
+		return -1, err
+	}
+
+	if err := m.setupswap(); err != nil {
+		fmt.Fprintln(os.Stdout, "Swap setup failed: ", err)
```
obbardc, line 656: just return err; no need to print

cmd/fakemachine/main.go
obbardc:
```
On-disk swap space size for the fakemachine (with a unit suffix, e.g. 2G); if unset, swap space will not be created
```

obbardc: & best to update the commit message to include some of these missing details, I think.

machine.go
```diff
@@ -452,6 +457,20 @@ func (m *Machine) SetScratch(scratchsize int64, path string) {
 	}
 }
 
+// SetSwap sets the size and location of on-disk swap space to allocate
```
obbardc, line 460: it'd be great to add this to the unit tests

cmd/fakemachine/main.go
evelikov: And by "these details" do you mean:
https://github.com/go-debos/fakemachine/pull/107#discussion_r981098545, or https://github.com/go-debos/fakemachine/pull/107#discussion_r981165743, or https://github.com/go-debos/fakemachine/pull/107#discussion_r981170122 ... or all three combined?

machine.go
evelikov: Fixed, thanks.

cmd/fakemachine/main.go
obbardc: the commit message is not detailled enough. It doesn't specify why the change is required.

Here's a good start:

```
machine: Add support for disk-backed swap space

Add support for enabling swap space on the machine which is backed by the host machine's disk.

This is required because ...

Swap space can be added to a fakemachine by specifying the `--swapsize` argument, e.g. `--swapsize=16G`

```

machine.go
evelikov: I'm thinking of adding `TestSwap` modelled after `TestMemory()`. Namely: set the swap, parse `/proc/swaps` and error out.
If you'd like anything else please let me know.

obbardc: sounds great to me.

cmd/fakemachine/main.go
sjoerdsimons: fwiw it's still not entirely clear to me why this is needed; Are your image builds really using more then 2G memory in the debos vm? dracut using `/tmp` doesn't really explain it as that's a tmpfs which can only use 50% of ram by default?

evelikov: I've observed intermittent CI failures during the dracut stage, when either 2G or 4G memory was used. Shortly before the failures started happening, we switched to using compressed linux-firmware files - where the Arch package for those was huge 1G or so.

The option was proposed as a way to fix those, personally I didn't debug the exact details.

Although ... looking at the git log, I cannot see this being deployed and our CI was disabled (due to unrelated reasons) around the time this was proposed.

So I'm not sure what's happening here - @santoshmahto7 can you shed some light?

santoshmahto7: > fwiw it's still not entirely clear to me why this is needed; Are your image builds really using more then 2G memory in the debos vm? dracut using `/tmp` doesn't really explain it as that's a tmpfs which can only use 50% of ram by default?

Yes image build (dracut stage) is taking more than 2G, sometime even crossing 4G and we are seeing OOM inside fakemachine. fakemachine is ramdisk based, so any disk operation(other than chrooted) is basically ram operation. There is always a chances of crossing ram limit. We cannot increase ram if host system has limitation on ram.

Even if dracut using ramfs as /tmp directory we need to have no limit of /tmp directory, otherwise "no space left on device" will be thrown and dracut will fail and that can be possible only  when we have swap space available inside fakemachine

evelikov: I think we've moved the runners from 4GB machines to 16/32GB ones, so I think we should do that. Nevertheless the option makes sense IMHO.

santoshmahto7: Yeah having sufficient memory for fakemachine will mitigate our dracut issue, but  I do see this patch as a feature rather than a fix for any issue, providing the same advantage as any machine having swap support.

evelikov: @obbardc @sjoerdsimons please let us know your take: Should we polish and merge or just abandon this IMHO useful feature?

obbardc: If you think it is a useful feature then I am not going to block merging it ;-). It doesn't affect the "core functionality" so IMO it's fine to add this feature.

But it needs to be properly cleaned up and another review round before I can merge it.

cmd/fakemachine/main.go
```diff
@@ -17,6 +17,7 @@ type Options struct {
 	Memory      int               `short:"m" long:"memory" description:"Amount of memory for the fakemachine in megabytes"`
 	CPUs        int               `short:"c" long:"cpus" description:"Number of CPUs for the fakemachine"`
 	ScratchSize string            `short:"s" long:"scratchsize" description:"On-disk scratch space size (with a unit suffix, e.g. 4G); if unset, memory backed scratch space is used"`
```
obbardc, line 19: need to update README.md with new arguments too

.gitignore
evelikov: Done, see https://github.com/go-debos/fakemachine/pull/120

evelikov:
Believe that everything should be addressed now. @obbardc can you give it another look, thanks

sjoerdsimons:
I fail to see the real use-case for this; fakemachine already had a `-m`  option to make the available memory bigger ; If the hosts machines memory isn't enough for whatever reason (which seems questionable in itself), then you can add a swapfile on the host instead of adding complexity to fakemachine.

PR #109, opened 2022-09-27
SetupEnviron: Fix golint warning

obbardc:
Fix the golint warnings.


evelikov:
\o/ thank you

sjoerdsimons (changes requested):
Can you make your commit message meaningful rather then just "fix golint warning"

sjoerdsimons:
> Can you make your commit message meaningful rather then just "fix golint warning"

btw this doesn't fix the warnings it changes the behaviour as previously we'd (unintentionally) just always call setEnvironment; If you want to fix purely the warnings we can just drop the if block :)

PR #112, opened 2022-10-04
Bringup user-mode-linux backend on Arch Linux

obbardc:
With https://github.com/go-debos/test-containers/pull/12 we
added support for User Mode Linux in the Arch Linux containers.
Enable the tests during CI runs.

(draft until https://github.com/go-debos/test-containers/pull/12 is merged)

### DIFF
--- a/backend_uml.go
+++ b/backend_uml.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"strings"
 
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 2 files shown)


backend_uml.go
```diff
@@ -10,6 +10,7 @@ import (
 	"os"
```
evelikov, line 10: IMHO the commit message would be cleaner if it covers the approach not the result. Namely:

On Arch a local version "-usermodelinux" is used, thus the modules are installed in `/usr/lib/modules/$version-usermodelinux". Effectively side-by-side to the normal modules.

backend_uml.go
```diff
@@ -62,23 +63,63 @@ func (b umlBackend) KernelPath() (string, error) {
 
 func (b umlBackend) ModulePath() (string, error) {
 	// make sure the UML modules exist
-	// on non-merged usr systems the modules still reside under /usr/lib/uml
+
+	/* On debian systems, the module reside under the first child directory under
+	 * /usr/lib/uml/modules; on non-merged usr systems the modules still reside
```
evelikov, line 68: Do we need the the merged vs non-merged comment - why aren't we using "mergedUsrSystem()" within the codebase?

backend_uml.go
```diff
@@ -62,23 +63,63 @@ func (b umlBackend) KernelPath() (string, error) {
...
+		}
+
+		return moddir, nil
 	}
-	if len(modSubdirs) != 1 {
-		return "", fmt.Errorf("could not determine which user-mode-linux modules to use")
+
+	/* On Arch systems, however, the modules are stored under /usr/lib/modules/
+	 * with a suffix, "-usermodelinux".
+	 */
```
evelikov, line 93: Out of curiosity: what does Fedora do about the modules? Looking around shows absolutely nothing.

backend_uml.go
```diff
@@ -50,11 +50,14 @@ func (b umlBackend) KernelRelease() (string, error) {
 
 func (b umlBackend) KernelPath() (string, error) {
 	// find the UML binary
-	kernelPath, err := exec.LookPath("linux.uml")
-	if err != nil {
-		return "", fmt.Errorf("user-mode-linux not installed")
+	kernelBinaries := []string{"linux.uml", "vmlinux"}
```
evelikov, line 54: Cannot see anything custom in the Arch recipe. On the other hand the [Debian one](https://salsa.debian.org/uml-team/user-mode-linux/-/blob/master/debian/rules#L114) is responsible for the different name and module location.

So I would recommend tweaking this and the previous patch to:
 - search the standard location first
 - tweak the commit message (+ potentially add inline comment) that Debian moves/renames things around

evelikov:
The documentation suggests that "Closes" is used for issues, not PRs. Since you're happy to move this forward, I'd say close it manually.

Also: it if were me, I would split the two unrelated test commits to a separate PR.

obbardc:
> The documentation suggests that "Closes" is used for issues, not PRs. Since you're happy to move this forward, I'd say close it manually.
> 
> Also: it if were me, I would split the two unrelated test commits to a separate PR.

I did both of these actions; the remaining open review comments are still outstanding :-)

obbardc:
We don't really care about the user-mode Linux backend anymore so we should close this to avoid bitrot.

What do you think @evelikov @evelikov-work ?

PR #115, opened 2022-10-24
Ensure systemd-resolved binary is available at runtime

obbardc:
Debian Unstable (and Fedora) has split systemd-resovled into a separate package which enables systemd-resolved as the default name resolver at install-time. Because of this, we cannot simply depend on systemd-resolved in the distro packaging as this may cause the default name resolver to change to systemd-resolved for some users.
    
If the user does not have system-resolved installed, fakemachine will currently fails with any action which requires hostname resolution:

    2022/09/19 10:49:09 ==== debootstrap ====
    2022/09/19 10:49:09 Debootstrap | W: Unable to read /etc/apt/apt.conf.d/ - DirectoryExists (2: No such file or directory)
    2022/09/19 10:49:09 Debootstrap | W: Unable to read /etc/apt/apt.conf.d/ - DirectoryExists (2: No such file or directory)
    2022/09/19 10:49:09 Debootstrap | I: Target architecture can be executed
    2022/09/19 10:49:09 Debootstrap | I: Retrieving InRelease
    2022/09/19 10:49:09 Debootstrap | I: Retrieving Release
    2022/09/19 10:49:09 Debootstrap | E: Failed getting release file http://deb.debian.org/debian/dists/stable/Release
    2022/09/19 10:49:09 debootstrap.log | amd64: ok
    2022/09/19 10:49:09 debootstrap.log | wget: unable to resolve host address 'deb.debian.org'
    2022/09/19 10:49:09 debootstrap.log | wget: unable to resolve host address 'deb.debian.org'

Instead of forcing the user to install systemd-resolved, detect whether the user has the binary installed and bubble up an error earlier if the binary has not been found.
    
See: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1020288
See: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1020690

--
Requires https://github.com/go-debos/test-containers/pull/14 to be merged first.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -662,6 +662,11 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 		return -1, err
 	}
 
+	/* Ensure systemd-resolved is available */
+	if _, err := os.Stat("/lib/systemd/systemd-resolved"); err != nil {
+		return -1, err
+	}
+
 	/* Amd64 dynamic linker */
 	err = w.CopyFile("/lib64/ld-linux-x86-64.so.2")
 	if err != nil {


machine.go
```diff
@@ -652,6 +652,7 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 
 	/* Add additional binary locations to $PATH */
 	os.Setenv("PATH", os.Getenv("PATH")+":"+prefix+"/sbin")
+	os.Setenv("PATH", os.Getenv("PATH")+":"+prefix+"/lib/systemd")
```
sjoerdsimons, line 655: /lib/systemd shouldn't ever be in the PATH

machine.go
```diff
@@ -663,6 +664,11 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 		return -1, err
 	}
 
+	/* Ensure systemd-resolved binary is available */
+	if _, err := exec.LookPath("systemd-resolved"); err != nil {
```
sjoerdsimons, line 668: systemd-reslolved isn't expected to be in the path really; we hardcode `/lib/systemd` elsewhere so you can just check if `/lib/systemd/systemd-resolved` exists (and optionally if it's executable) which will work on both merged and non-merged user systems

machine.go
```diff
@@ -652,7 +650,10 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 		prefix = "/usr"
 	}
 
-	// search for busybox; in some distros it's located under /sbin
+	/* Add additional binary locations to $PATH */
+	os.Setenv("PATH", os.Getenv("PATH")+":"+prefix+"/sbin")
```
sjoerdsimons, line 654: In a non-merged usr setup you want to have both paths added to the path, not just the one. On a merged-usr system they collapse into one thing which is fine. So i'd suggest simply dropping this change entirely as it doesn't have a benefit and has the potential to cause regressions

sjoerdsimons (changes requested):

machine.go
obbardc: right I checked there first and fallback to checking $PATH

machine.go
```diff
@@ -662,6 +662,15 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 		return -1, err
 	}
 
+	/* Ensure systemd-resolved is available: Debian/Arch ship under /lib/systemd
+	 * while Fedora (and other distros) ship under /usr/bin/
+	 */
+	if _, err := os.Stat("/lib/systemd/systemd-resolved"); err != nil {
+		if _, err := exec.LookPath("systemd-resolved"); err != nil {
```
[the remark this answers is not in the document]
obbardc: right i just checked again and it seems that fedora puts it in /usr/bin and /lib/systemd; go figure ;-)

sjoerdsimons:
Fwiw the debian bugs; the right solution ther eis to update the Debian package with the dependency, these changes are more relevant for installations from source

PR #117, opened 2022-10-24
Add a man page

obbardc:
During the latest Debian package update, lintian warned that no manual page for fakemachine exists. Add a manual page so that the next import closes this warning.


evelikov:
What's the reason behind having the generated artefacts in git? I've seen many projects avoid that and only a few that do.

obbardc:
> What's the reason behind having the generated artefacts in git? I've seen many projects avoid that and only a few that do.

Really at this point it's to have unity with debos and get a man page generated. It's one less thing to do at release time too. I have no preference ;-).

sjoerdsimons:
would be nice to improve and not have the man page in git in future at least; but ok for now

obbardc:
Bors retry

obbardc:
bors retry

PR #119, opened 2022-10-30
Allow to set kernel on machine level

avoidik:


avoidik:
this will allow us to set custom kernel and initrd on machine level like this

```go
m, err = fakemachine.NewMachineWithBackend(options.Backend)
m.SetMemory(int(2097152 / 1024 / 1024))
m.SetNumCPUs(2)
m.SetKernelPath("vmlinuz")
m.SetInitrdPath("initrd.gz")
m.SetShowBoot(true)
```

avoidik:
Most probably external control on `m.initrdpath` in this context is not really needed, since it's a temporary file

machine.go
```diff
@@ -458,6 +459,14 @@ func (m *Machine) SetScratch(scratchsize int64, path string) {
 	}
 }
 
+func (m *Machine) SetInitrdPath(path string) {
```
obbardc, line 462: I don't see the benefit of changing the initrd path since it's generated on demand by fakemachine?

avoidik: I agree, let remove it

cmd/fakemachine/main.go
```diff
@@ -18,6 +19,7 @@ type Options struct {
 	CPUs        int               `short:"c" long:"cpus" description:"Number of CPUs for the fakemachine"`
 	ScratchSize string            `short:"s" long:"scratchsize" description:"On-disk scratch space size (with a unit suffix, e.g. 4G); if unset, memory backed scratch space is used"`
 	ShowBoot    bool              `long:"show-boot" description:"Show boot/console messages from the fakemachine"`
+	KernelPath  string            `short:"k" long:"kernel" description:"kernel to use"`
```
evelikov, line 22: When you say "kernel to use", is that just the filename (within the usual search paths), the complete absolute path + filename, or the path relative to `RECIPE_DIR` + filename?

Please enhance the help message so it's clear which one we're talking about. Also do update the README.md thanks o/

obbardc, line 22: as stated in my review of https://github.com/go-debos/debos/pull/376 it'd be much better to implement this as a `backed-option`, so something like:

`--backend-option kernel-path:kernel.img`

```
if err = m.SetBackendOption("key", "value"); err != nil {
    // show error to user
}
```

sjoerdsimons:
@avoidik i assume you're no long interested in this? I'll close this for now, please feel free to re-open once you get back to it

PR #129, opened 2023-01-24
Add arm64 support & fixes for running on /usr-merged/Fedora systems

refi64:
For the arm64 support, this uses a very different approach from #30, but still generalizes things enough to add support for other architectures in the future. Tested to build an Apertis image from a Fedora host.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -1,5 +1,5 @@
-//go:build linux && amd64
-// +build linux,amd64
+//go:build linux
+// +build linux
 
 package fakemachine
 
--- a/backend_qemu.go
+++ b/backend_qemu.go
@@ -1,5 +1,5 @@
-//go:build linux && amd64
-// +build linux,amd64
+//go:build linux
+// +build linux
 
 package fakemachine
 
@@ -35,8 +35,18 @@ func (b qemuBackend) Supported() (bool, error) {
 	return true, nil
 }
 
+var qemuBinaryNames = map[Arch]string{
+	Amd64: "qemu-system-x86_64",
+	Arm64: "qemu-system-aarch64",
+}
+
 func (b qemuBackend) QemuPath() (string, error) {
-	return exec.LookPath("qemu-system-x86_64")
+	binary, ok := qemuBinaryNames[b.machine.arch]
+	if !ok {
+		return "", fmt.Errorf("unsupported arch for qemu: %s", b.machine.arch)
+	}
+
+	return exec.LookPath(binary)
 }
 
 func (b qemuBackend) KernelRelease() (string, error) {
@@ -173,7 +183,7 @@ func (b qemuBackend) StartQemu(kvm bool) (bool, error) {
 	}
 	memory := fmt.Sprintf("%d", m.memory)
 	numcpus := fmt.Sprintf("%d", m.numcpus)
-	qemuargs := []string{"qemu-system-x86_64",
+	qemuargs := []string{qemuBinaryNames[m.arch],
 		"-smp", numcpus,
 		"-m", memory,
 		"-kernel", kernelPath,
@@ -185,6 +195,14 @@ func (b qemuBackend) StartQemu(kvm bool) (bool, error) {
 		qemuargs = append(qemuargs,
 			"-cpu", "host",
 			"-enable-kvm")
+	}
+
+	if m.arch == Arm64 {
+		qemuargs = append(qemuargs, "-machine", "virt")
+		if !kvm {
+			// QEMU uses a 32-bit CPU by default, so override it.
+			qemuargs = append(qemuargs, "-cpu", "cortex-a53")
+		}
 	}
 
 	kernelargs := []string{"console=ttyS0", "panic=-1",
--- a/machine.go
+++ b/machine.go
@@ -681,7 +681,7 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 	}
 
 	/* Ensure systemd-resolved is available */
-	if _, err := os.Stat("/lib/systemd/systemd-resolved"); err != nil {
+	if _, err := os.Stat(prefix + "/lib/systemd/systemd-resolved"); err != nil {
 		return -1, err
 	}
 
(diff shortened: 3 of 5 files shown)


machine.go
```diff
@@ -667,21 +685,39 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 		return -1, err
 	}
 
-	/* Amd64 dynamic linker */
-	err = w.CopyFile("/lib64/ld-linux-x86-64.so.2")
+	var dynamicLinker string
+	if m.arch == Arm64 {
```
obbardc, line 697: would it be worth adding a table of arch vs. path for these, and erroring out if it's not a supported arch (like we do for Qemu path for instance) ?

backend_qemu.go
```diff
@@ -187,6 +197,14 @@ func (b qemuBackend) StartQemu(kvm bool) (bool, error) {
 			"-enable-kvm")
 	}
 
+	if m.arch == Arm64 {
```
obbardc, line 200: did you test this with both qemu and kvm backends ?

machine.go
```diff
@@ -667,21 +685,39 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
...
 
 	/* C libraries */
-	libraryDir, err := realDir("/lib64/ld-linux-x86-64.so.2")
+	libraryDir, err := realDir(dynamicLinker)
 	if err != nil {
 		return -1, err
 	}
+
 	err = w.CopyFile(libraryDir + "/libc.so.6")
+	if err != nil && errors.Is(err, os.ErrNotExist) && m.arch == Arm64 && strings.HasSuffix(libraryDir, "/lib") {
```
obbardc, line 726: would it be worth making the check for arch more clear ?
or even, do we need the arch check here ?

machine.go
```diff
@@ -685,23 +685,31 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
...
 
-	/* dynamic linker */
-	err = w.CopyFile(dynamicLinker)
+	/* dynamic linkers */
+	err = w.CopyFile(glibcDynamicLinker)
 	if err != nil {
 		return -1, err
 	}
 
+	// Fedora's busybox needs musl
```
obbardc, line 713: nit: keep comment style of `/* */`
also worth saying that muscl is optional here, it's not really clear from a first look :-)

obbardc (changes requested):
Small nits really, thank you for doing this work!

obbardc:
It'd also be great to add tests for Fedora on amd64. See https://github.com/go-debos/test-containers/issues/9 for more information.

I don't think we can do tests on arm64, but we could at least crossbuild for that arch?

backend.go
```diff
@@ -1,5 +1,5 @@
-//go:build linux && amd64
-// +build linux,amd64
+//go:build linux
+// +build linux
```
evelikov, line 2: I'm really happy to see arm64 support. Thank you. Did you try it with Fedora only or did you also give Debian/Arch a try?

Small nit: Personally, I would add the new/supported arch instead of removing the existing restriction.

With the proposed change, one can trivially build binaries that are known broken. Which in itself is poor user experience IMHO.

obbardc: I think that's fine as we could also add support for currently unsupported archs ;-) - as long the error is explicit at runtime I am happy.

backend_qemu.go
```diff
@@ -173,7 +183,7 @@ func (b qemuBackend) StartQemu(kvm bool) (bool, error) {
 	}
 	memory := fmt.Sprintf("%d", m.memory)
 	numcpus := fmt.Sprintf("%d", m.numcpus)
-	qemuargs := []string{"qemu-system-x86_64",
+	qemuargs := []string{qemuBinaryNames[m.arch],
```
obbardc, line 186: why not call `QemuPath()` here ?

obbardc (changes requested):
It'd also be great to get a CI test for this by creating a fedora test container and hooking that into CI: https://github.com/go-debos/test-containers/issues/9

sjoerdsimons:
Fwiw did #157  based on heavily refactoring your first patch to make it more generic and fully based on a per-architecture configuration rather then in-code exceptions (apart from the qemu binary).. Also compared to this has a working  `--show-boot` on arm :)

machine.go
sjoerdsimons: This needs a more generic fix to get this to work on fedora.

The overall assumption this code makes is that the main library directory *matches* the real location of the dynamic linker.

This is true  for debian where the canonical linker locations are  e.g. "/lib/ld-linux-aarch64.so.1" or "/lib64/ld-linux-x86-64.so.2" are symlinks to "
aarch64-linux-gnu/ld-linux-aarch64.so.1" and "/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2" respectively.. 

It also works for arch as it only seems to use `/lib64` on amd64 where it's just a symlink to `/usr/lib` anyway.

So fedora is the odd one out here in that it uses `/usr/lib64`  on arm64 *as well* and having it's dynamic linkers  directly in their canonical location.. Which causes the failure on arm64 as the linker in `lib` but everything else in `lib64`

TL;DR to cover fedora (and the likes) properly the way we determine libraryDir should change/be expanded, and probably broken out to an helper, rather then patching it up after the fact

machine.go
```diff
@@ -217,10 +217,18 @@ func NewMachineWithBackend(backendName string) (*Machine, error) {
...
 	}
+	if _, err := os.Stat("/etc/pki"); err == nil {
+		m.AddVolume("/etc/pki")
+	}
 	if _, err := os.Stat("/etc/ssl"); err == nil {
 		m.AddVolume("/etc/ssl")
 	}
 
+	// Mounts for gnutls (used by wget).
+	if _, err := os.Stat("/etc/crypto-policies"); err == nil {
```
sjoerdsimons, line 228: crypto-policies is not just for gnutls it's the centralized policies for all crypto :)

sjoerdsimons:
given how long this has been stale and now conflicting i'll close it for now; @refi64 please feel free to re-open once you get back to it :)

PR #141, opened 2023-04-05
Wrap errors where possible and makes sense

andrewshadura:
Using `%w` instead of `%v` makes `Errorf` store the original error, allowing the caller to extract and use it later. While there isn’t any immediate benefit from it now, it may be useful to the users of fakemachine.

See:
 * https://go.dev/doc/go1.13#error_wrapping
 * https://github.com/golang/go/wiki/ErrorValueFAQ
 * https://go.googlesource.com/proposal/+/master/design/29934-error-values.md

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -833,7 +833,7 @@ func (m *Machine) startup(command string, extracontent [][2]string) (int, error)
 
 	success, err := backend.Start()
 	if !success || err != nil {
-		return -1, fmt.Errorf("error starting %s backend: %v", backend.Name(), err)
+		return -1, fmt.Errorf("error starting %s backend: %w", backend.Name(), err)
 	}
 
 	result, err := os.Open(path.Join(tmpdir, "result"))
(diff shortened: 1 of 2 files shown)


obbardc:
Does this change the output of the API consumers in any way? It'd be good to have clarification of that before merging.

machine.go
```diff
@@ -867,7 +867,7 @@ func (m *Machine) RunInMachineWithArgs(args []string) (int, error) {
 	executable, err := exec.LookPath(os.Args[0])
 
 	if err != nil {
-		return -1, fmt.Errorf("Failed to find executable: %v\n", err)
+		return -1, fmt.Errorf("Failed to find executable: %w\n", err)
```
obbardc, line 870: Seems like the `\n` is redundant & also the error message shouldn't start with a capital (not introduced here of course)

andrewshadura:
> Does this change the output of the API consumers in any way? It'd be good to have clarification of that before merging.

According to [this](https://github.com/golang/go/wiki/ErrorValueFAQ#how-can-i-add-context-to-an-error-im-already-returning-without-breaking-clients), it should not.

machine.go
obbardc: I will fix this in a follow-up.

obbardc (approved):

PR #160, opened 2023-07-26
backend: Rework newBackend() error handling

obbardc:
The newBackend() function may enumerate multiple backends to find the best supported backend for the system. Because of this, the function may return multiple errors.

There are two problems with how this function returns multiple errors:
 - if only one error occurs, the error is appended to a nil error
 - they are wrapped with %v instead of %w

Correct these issues.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -51,7 +51,12 @@ func newBackend(name string, m *Machine) (backend, error) {
 
 			b, backendErr := newBackend(backendName, m)
 			if backendErr != nil {
-				err = fmt.Errorf("%v, %v", err, backendErr)
+				/* Append the error to existing error */
+				if err != nil {
+					err = fmt.Errorf("%w, %w", err, backendErr)
+				} else {
+					err = fmt.Errorf("%w", backendErr)
+				}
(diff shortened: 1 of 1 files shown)


backend.go
```diff
@@ -51,7 +51,12 @@ func newBackend(name string, m *Machine) (backend, error) {
 
 			b, backendErr := newBackend(backendName, m)
 			if backendErr != nil {
-				err = fmt.Errorf("%v, %v", err, backendErr)
+				/* Append the error to existing error */
+				if err != nil {
+					err = fmt.Errorf("%w, %w", err, backendErr)
```
obbardc, line 56: seems like two errors cannot be appended like this, hm

sjoerdsimons: that;s suprirsing; though mind you'd probably also want to put the backend name per error

obbardc:
Handling in https://github.com/go-debos/fakemachine/pull/160

sjoerdsimons:
> Handling in #160

Not sure what you mean with that; since #160 is this PR?

PR #162, opened 2023-07-26
Disable plymouth boot splash

obbardc:
If plymouth is installed in the host, the service  will be started during
the fakemachine's boot sequence. When plymouth starts, it attempts to
clear all of the the serial consoles which in turn causes the host's
shell to be cleared.

If fakemachine is ran with --show-boot this causes the boot log to be
lost and can also cause strange artifacts on the shell (e.g. missing
characters).

Add plymouth.enable=0 to the kernel arguments to stop plymouth from
starting during the machine's boot. The uml backend has the quiet parameter
which was added to achieve the same goal, so unify the kernel arguments
between the backends.

### DIFF
--- a/backend_qemu.go
+++ b/backend_qemu.go
@@ -224,6 +224,7 @@ func (b qemuBackend) StartQemu(kvm bool) (bool, error) {
 	qemuargs = append(qemuargs, "-machine", qemuMachine.machine)
 	console := fmt.Sprintf("console=%s", qemuMachine.console)
 	kernelargs := []string{console, "panic=-1",
+		"plymouth.enable=0",
 		"systemd.unit=fakemachine.service"}
 
 	if m.showBoot {
--- a/backend_uml.go
+++ b/backend_uml.go
@@ -205,7 +205,7 @@ func (b umlBackend) Start() (bool, error) {
 		"mem=" + memory + "M",
 		"initrd=" + m.initrdpath,
 		"panic=-1",
-		"nosplash",
+		"plymouth.enable=0",
 		"systemd.unit=fakemachine.service",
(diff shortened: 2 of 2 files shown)


backend_qemu.go
```diff
@@ -224,8 +224,11 @@ func (b qemuBackend) StartQemu(kvm bool) (bool, error) {
 	qemuargs = append(qemuargs, "-machine", qemuMachine.machine)
 	console := fmt.Sprintf("console=%s", qemuMachine.console)
 	kernelargs := []string{console, "panic=-1",
+		"nosplash",
 		"systemd.unit=fakemachine.service"}
 
+	kernelargs = append(kernelargs, "nosplash")
+
```
evelikov, line 231: I'm seeing double? Seems like we're adding the argument twice.

Should this be dependant on the "show-boot" option? In the default use-case I don't want to see any kernel/boot log, which plymouth hides nicely.

evelikov:
Food for thought after briefly trying plymouth on one my systems:

It seems like `mutl-user.target.wants` pulls `plymout-quit{,-wait}.service` which ensures we don't get into such issues. Not sure if fakemachine is usin multi-user ... perhaps it should? If that's not an option quick search suggests that `rd.plymouth=0 plymouth.enable=0` tends to be the better option than `nosplash`.

Either way, HTH

obbardc:
> a quick search suggests that `rd.plymouth=0 plymouth.enable=0` tends to be the better option than `nosplash`.

@evelikov I updated the PR to use this parameters and tested across all combinations of backends and `--show-boot` on my Debian sid system. Can you check on Arch, please?

Issue #194, opened 2024-02-12
fakemachine fails on debian kernel version 6.6.13 because kernel modules are now xz compressed


obbardc:
fakemachine fails with:

```
[    1.099950] Run /init as init process
modprobe: can't load module virtio_console (/usr/lib/modules/6.6.13-amd64/kernel/drivers/char/virtio_console.ko): No such file or directory
modprobe: can't load module 9pnet (/usr/lib/modules/6.6.13-amd64/kernel/ne[   t 1.113319/] Kerne9l ppanic -/ not sync9ing: pAttempted to knill init! exitcode=0x00007f00
```

This is because the kernel modules are now xz compressed e.g: `/usr/lib/modules/6.6.13-amd64/kernel/net/9p/9pnet.ko.xz`

This change happened in `6.6.3-1~exp1`;  https://salsa.debian.org/kernel-team/linux/-/commit/e8b64e4b4318aaaf7222f47922c8dfe91b1dd869

obbardc:
So we do handle compressed modules; but I believe we currently attempt to extract them into the created initrd.

I think this was because busybox's modprobe didn't handle loading compressed kernel modules out-of-the box?

obbardc:
OK seems like there are two issues here:


1) module copying/extraction code is flawed https://github.com/go-debos/fakemachine/blob/377f277b4c55fa867544335a6cef617db96518a5/machine.go#L119

We prepend `/usr` and essentially extract e.g. `/usr/lib/modules/6.6.13-amd64/kernel/fs/netfs/netfs.ko.xz` to `/usr/usr/lib/modules/6.6.13-amd64/kernel/fs/netfs/netfs.ko`


2) We extract the module into the initrd, but the new Debian kernel doesn't seem to expect that:
```
$ find /usr/lib/modules
/usr/lib/modules/6.6.13-amd64
/usr/lib/modules/6.6.13-amd64/modules.builtin
/usr/lib/modules/6.6.13-amd64/modules.alias
/usr/lib/modules/6.6.13-amd64/modules.symbols
/usr/lib/modules/6.6.13-amd64/kernel
/usr/lib/modules/6.6.13-amd64/kernel/net
/usr/lib/modules/6.6.13-amd64/kernel/net/9p
/usr/lib/modules/6.6.13-amd64/kernel/net/9p/9pnet_virtio.ko
/usr/lib/modules/6.6.13-amd64/kernel/net/9p/9pnet.ko
[...]

PR #195, opened 2024-02-12
Fix compatibility when kmod module_directory prefixed with /usr

obbardc:
Debian's kmod now builds with `--with-module-directory=/usr/lib/modules`[1]
(while the default is `--with-module-directory=/lib/modules`) which means
that `modinfo` now outputs module paths with `/usr` prefix already.

In fakemachine, on merged-usr systems, modules are copied into the initrd
with a prefixed path of `/usr`, no matter the system path produced by modinfo.
This means that modules are now copied into the wrong place, e.g /usr/usr/....

Fix compatibility with kmod built with /usr prefixed module-directory by
ensuring the module destination path has a prefix of /usr, rather than
always prefixing the destination path with /usr.

[1]: https://salsa.debian.org/md/kmod/-/commit/3306fbb10a8ff6c3bd643437e8f29353872524fd
Fixes: https://github.com/go-debos/fakemachine/issues/194

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -99,11 +99,6 @@ func (m *Machine) copyModules(w *writerhelper.WriterHelper, modname string, copi
 		return nil
 	}
 
-	prefix := ""
-	if mergedUsrSystem() {
-		prefix = "/usr"
-	}
-
 	found := false
 	for suffix, fn := range suffixes {
 		if strings.HasSuffix(modpath, suffix) {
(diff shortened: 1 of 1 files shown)


sjoerdsimons (approved):
Fwiw you commit message is a bit misleading; There is no change in system path,  what you're seeing the effect of kmod now having `/usr/` as the default searchpath it seems as per https://salsa.debian.org/md/kmod/-/commit/3306fbb10a8ff6c3bd643437e8f29353872524fd

obbardc:
Argh so yeah I just tried older kmod and things work fine. Updated commit message to be less confusing ;-).

sjoerdsimons:
> Argh so yeah I just tried older kmod and things work fine. Updated commit message to be less confusing ;-).

If i'm not mistaken you only chagned the title of this PR not the patch commit message ?

obbardc:
@sjoerdsimons I didn't push.

sjoerdsimons (approved):

Issue #207, opened 2024-05-13
Fails to provide network on IPv6-only hosts


helmutg:
Running fakemachine on an IPv6-only host results in "Network is unreachable" when trying to access the network. The network stack usually tries IPv6 first, but `machine.go` sets `LinkLocalAddressing=no` and `IPv6AcceptRA=no`, so it does not have any IPv6 address nor route and the Linux kernel inside KVM reports `-ENETUNREACH`. It then falls back to IPv4 and sends a packet. KVM's slirp-based user network stack then tries to send this packet, but observes that the host machine does not have any IPv4 addresses nor routes and generates an ICMPv4 packet indicating `-ENETUNREACH`. Having failed both address families, `-ENETUNREACH` is reported to the end user. This makes e.g. `debos` fail.

Debian-Bug: https://bugs.debian.org/1071033

obbardc:
@sjoerdsimons do you have any ideas about this ?

PR #208, opened 2024-05-14
enable network on ipv6-only hosts

helmutg:
The networkd template explicitly disables IPv6-connectivity. When connecting to a host, the attempt to use IPv6 results in -ENETUNREACH from the guest kernel. If the host is IPv6-only, the host kernel likewise makes attempts to use IPv4 result in -ENETUNREACH. Hence fakemachine has dysfunctional network when invoked on an IPv6-only host.

Closes: #207

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -337,8 +337,8 @@ Description=fakemachine runner
 Conflicts=shutdown.target
 Before=shutdown.target
 Requires=basic.target
-Wants=systemd-resolved.service binfmt-support.service systemd-networkd.service
-After=basic.target systemd-resolved.service binfmt-support.service systemd-networkd.service
+Wants=systemd-resolved.service binfmt-support.service network-online.target
+After=basic.target systemd-resolved.service binfmt-support.service network-online.target
 OnFailure=poweroff.target
 
 [Service]
--- a/machine.go
+++ b/machine.go
@@ -301,9 +301,6 @@ Type=ether
 
 [Network]
 DHCP=ipv4
-# Disable link-local address to speedup boot
-LinkLocalAddressing=no
-IPv6AcceptRA=no
 `
 
 const networkdLinkTemplate = `


machine.go
```diff
@@ -301,9 +301,6 @@ Type=ether
 
 [Network]
 DHCP=ipv4
-# Disable link-local address to speedup boot
```
obbardc, line 304: Does this cause any slow-down as per the comment above ?

machine.go
```diff
@@ -301,9 +301,6 @@ Type=ether
 
 [Network]
 DHCP=ipv4
```
obbardc, line 303: does this work on networks with ipv6 ?

machine.go
helmutg: It quite definitely causes a slow-down. Not sure how much. To me the question is more of whether it works at all. I cannot use `debos` at all as a result of this problem.

machine.go
helmutg: Yes. For IPv6, we typically do not use DHCP but ICMPv6 based autoconfiguration. We need a link local IPv6 address to perform this auto configuration and we need to accept router advertisements to configure a gateway.

helmutg: Maybe your question really was whether it also works on networks with IPv4 or with both IPv4 and IPv6. In those cases, DHCP works for IPv4 as before and the fakemachine may have dual stack network connectivity preferring IPv6.

machine.go
obbardc: Sorry for the late reply - I've been too busy to do any debos maintainance. Can you check if adding `MaxAttempts=10` (with the original `LinkLocalAddressing=no`) and possibly `IPv6AcceptRA=yes` works for ipv6-only ?

sjoerdsimons: LinkLocal=no with IPv6AcceptRA seems wrong; Just doing ipv6 link local addressing would make more sense; 

Mind ofcourse this change was done in 2017 so things in systemd might have changed quite a bit.  In particular the behaviour of networkd-wait-online could well have changed to not wait for all address families (which i *think* was the reason for the slowdown) 

@helmutg when you say definitely a slowdown; Is that something that you actually noticed during usage or more an expectation :)

helmutg: Sorry for the delay.

I cannot actually observe any delay with these settings, because `fakemachine` does not actually work in any way unless I change them. It is only that I expect a slowdown.

So I did some more testing. With both `IPv6AcceptRA=yes` and `LinkLocalAddressing=yes`, it works most of the time. I had at least one failure and suspect that `fakemachine.service` was being run before the network interface succeeded in configuring. Just setting `MaxAttempts=10` does not improve the situation in any way.

Adding `MaxAttempts=10` and `IPv6AcceptRA=yes` also does not work. Neither `IPv6AccptRA=yes` nor `LinkLocalAddressing=yes` make it work in isolation. It is only when we combine them that it starts to work. Without the former, we don't get a default route and without the latter we don't get an IPv6 address assigned that could be routed.

In all of this, when I say "does not work", what I mean is "Network is unreachable" being part of the error message.

Given further debugging, I think this is not the full solution. Network interfaces are not reliably configured when `systemd-networkd.service` is started. What is really needed here is `network-online.target`, but that's not `WantedBy` or `Before` `fakemachine.service` in any way. Of course adding `network-online.target` will make things even slower (again not measured in any way, just an expectation that adding more dependencies will take longer). But then in the context I am trying to use it, it currently does not work at all (due to not supplying any working IPv4 connectivity).

sjoerdsimons: The command wrapper does run   `/lib/systemd/systemd-networkd-wait-online -q --interface=ethernet0` to wait for network connectivity, so adding network-online shouldn't be needed.  Looks like your change make our CI tests fail for some reason (which tbh is surprising).

helmutg: Your argument sounds reasonable to me, but we migrated away from debos towards debvm, which solves our niche use case way faster and more reliably (by not spinning up a secondary linux kernel). I'll leave things up to you here. If the MR stops being productive, please close it.

sjoerdsimons: No worries; I'll close it for now :) Thanks for the quick response!

PR #220, opened 2024-10-30
cmd: Add 'version' option

### LINKED ISSUE #200: add --version flag which prints version

daissi:
This 'version' option prints the fakemachine version. Fix #200

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,9 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v4
 
+    - name: Set safe directory in container
+      run: /usr/bin/git config --global --add safe.directory ${GITHUB_WORKSPACE}
+
     - name: Define fakemachine version
       run: echo "FAKEMACHINE_VER=$(git describe --tags HEAD)" >> "$GITHUB_ENV"
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,9 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        filter: tree:0
 
     - name: Set safe directory in container
       run: /usr/bin/git config --global --add safe.directory ${GITHUB_WORKSPACE}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,8 +73,14 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v4
 
+    - name: Define fakemachine version
+      run: echo "FAKEMACHINE_VER=$(git describe --tags HEAD)" >> "$GITHUB_ENV"
+
     - name: Test build
-      run: go build -o fakemachine cmd/fakemachine/main.go
+      run: go build -ldflags="-X main.Version=${FAKEMACHINE_VER}" ./cmd/fakemachine
+
+    - name: Print fakemachine version
+      run: ./fakemachine --version
 
     - name: Run unit tests (${{matrix.backend}} backend)
       run: go test -v ./... --backend=${{matrix.backend}} | tee test.out
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,7 +77,7 @@ jobs:
       run: /usr/bin/git config --global --add safe.directory ${GITHUB_WORKSPACE}
 
     - name: Define fakemachine version
-      run: echo "FAKEMACHINE_VER=$(git describe --tags HEAD)" >> "$GITHUB_ENV"
+      run: echo "FAKEMACHINE_VER=$(git describe --always --tags HEAD)" >> "$GITHUB_ENV"
 
     - name: Test build
       run: go build -ldflags="-X main.Version=${FAKEMACHINE_VER}" ./cmd/fakemachine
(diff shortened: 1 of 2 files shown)


obbardc (approved):

obbardc:
CI fails because of https://github.com/go-debos/fakemachine/pull/218

cmd/fakemachine/main.go
```diff
@@ -11,6 +11,8 @@ import (
 	"strings"
 )
 
+const Version = "0.0.10"
```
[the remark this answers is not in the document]
obbardc: Untested but pretty sure we can inject the version at build time with `go build -ldflags "-X main.Version=$VERSION"`.

How do you feel about
1) injecting the version in the GitHub action based on git commit
2) injecting the same version in the Debian packaging

I looked for existing libraries to do this, like in rust, but couldn't find anything.

(Also the same will be true for debos)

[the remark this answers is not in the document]
sjoerdsimons: Is that what's typical in the go space (surely others do this as well)? But yeah in principle that sounds fine, my main concern is really that it's too easy to forget to update :)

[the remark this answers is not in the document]
daissi: Now, the version is injected at build time based on `git describe --tags HEAD`. Is that better?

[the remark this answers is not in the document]
sjoerdsimons: It's getting there; but it's incomplete.. I ended up doing a bit of a deep dive on go versioning yesterday...

Essentially we've got 3 different variations:
1. Build where the build system can set a version;
1. self-build fakemachine from a remote repo e.g. `go install github.com/go-debos/fakemachine/cmd/fakemachine@latest` . 
1. self-build fakemachine from a local git checkout e.g. `go build ./cmd/fakemachine` 

This patch is only covering the first case, by taking advantage of our CI pipeline setting the version. The Debian packaging case can be covered the same way. For example the [caddy packaging](https://salsa.debian.org/go-team/packages/caddy/-/blob/debian/sid/debian/rules?ref_type=heads#L13)  ends up doing that.

For the other two cases it seems best to use the golang [debug buildinfo](https://pkg.go.dev/runtime/debug#BuildInfo). When building from a remote info the `Main` field ends up with a reasonable version, while when building locally the `Settings` field contains [vcs information](https://pkg.go.dev/runtime/debug#BuildSetting). 

as a POC i played with https://github.com/sjoerdsimons/go-version-test/blob/main/cmd/go-version-test/main.go, which seems to what i'd expect in the various different cases (see last [actions run](https://github.com/sjoerdsimons/go-version-test/actions)). So we probably want someting similar here

obbardc:
I've added this to the [v0.0.9](https://github.com/go-debos/fakemachine/milestone/4) milestone, the changes look small enough to me to get that in the next week or so? :-)

cmd/fakemachine/main.go
```diff
@@ -11,6 +11,8 @@ import (
 	"strings"
 )
 
+var Version string = "none"
```
obbardc, line 14: Does it make sense to expose this in the fakemachine go API (so we could e.g print the version of fakemachine used in debos debug) ? It could come later after this PR.

.github/workflows/ci.yml
```diff
@@ -74,7 +74,7 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Test build
-      run: go build -o fakemachine cmd/fakemachine/main.go
+      run: go build -o fakemachine -ldflags="-X main.Version=$(git describe --tags HEAD)" cmd/fakemachine/main.go
```
obbardc, line 77: I'd rather build the version as a variable (which could be overwritten?) first in a separate step and output as variable.

obbardc (changes requested):

sjoerdsimons (approved):

sjoerdsimons (changes requested):
see long comment :)

cmd/fakemachine/main.go
sjoerdsimons: That's very hard with golang; remember is a static library. Fakemachine, the library, itself has no idea what version it is inherently. Debos could figure it our from the golang buildinfo, but that gets cleared when doing debian builds for some reason

obbardc: Sounds reasonable enough.

sjoerdsimons:
> I've added this to the [v0.0.9](https://github.com/go-debos/fakemachine/milestone/4) milestone, the changes look small enough to me to get that in the next week or so? :-)

well i already tagged 0.0.11 :)  As i noted in one of the threads ; golang versioning is full of subtleties, so i expect it to at least take some review rounds

obbardc:
> > I've added this to the [v0.0.9](https://github.com/go-debos/fakemachine/milestone/4) milestone, the changes look small enough to me to get that in the next week or so? :-)
> 
> well i already tagged 0.0.11 :) As i noted in one of the threads ; golang versioning is full of subtleties, so i expect it to at least take some review rounds

ok, I set the correct milestone now.

.github/workflows/ci.yml
daissi: This should be fixed now

sjoerdsimons (approved):
Changes look good to me; Sadly CI fails but it's not clear why (doesn't look like your changes could cause it but...)

obbardc (approved):

PR #222, opened 2025-02-14
Add method to override the images sector size

julien-massot:
The sector size is added at drive level, so it's correctly reported by the kernel, allowing userland software to works without overriding the values. UML backend doesn't support alternate sector size since the ubd driver hardcode the 512 bytes sector size. (arch/um/drivers/ubd_kern.c)

### DIFF
--- a/backend_uml.go
+++ b/backend_uml.go
@@ -45,6 +45,11 @@ func (b umlBackend) Supported() (bool, error) {
 	if _, err := b.SlirpHelperPath(); err != nil {
 		return false, fmt.Errorf("libslirp-helper not installed")
 	}
+
+	// only support 512 bytes sector size
+	if b.machine.sectorSize != 512 {
+		return false, fmt.Errorf("uml backend only supports 512 bytes sector size")
+	}
 	return true, nil
 }
 
--- a/cmd/fakemachine/main.go
+++ b/cmd/fakemachine/main.go
@@ -18,6 +18,7 @@ type Options struct {
 	EnvironVars map[string]string `short:"e" long:"environ-var" description:"Environment variables (use -e VARIABLE:VALUE syntax)"`
 	Memory      int               `short:"m" long:"memory" description:"Amount of memory for the fakemachine in megabytes"`
 	CPUs        int               `short:"c" long:"cpus" description:"Number of CPUs for the fakemachine"`
+	SectorSize  int               `short:"S" long:"sectorsize" description:"Override image sector size"`
 	ScratchSize string            `short:"s" long:"scratchsize" description:"On-disk scratch space size (with a unit suffix, e.g. 4G); if unset, memory backed scratch space is used"`
 	ShowBoot    bool              `long:"show-boot" description:"Show boot/console messages from the fakemachine"`
 	Quiet       bool              `short:"q" long:"quiet" description:"Don't show logs from fakemachine or the backend; only print the command's stdout/stderr"`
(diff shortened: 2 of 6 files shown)


backend_uml.go
```diff
@@ -45,6 +45,11 @@ func (b umlBackend) Supported() (bool, error) {
 	if _, err := b.SlirpHelperPath(); err != nil {
 		return false, fmt.Errorf("libslirp-helper not installed")
 	}
+
+	// only support 512 bytes sector size
+	if b.machine.sectorSize != 512 {
+		return false, fmt.Errorf("uml backend only supports 512 bytes sector size.")
```
obbardc, line 51: nit: remove dot at end of line.

sjoerdsimons:
Would be great to have a test as well, but that can come later

backend_uml.go
julien-massot: Thanks removed

julien-massot:
> Would be great to have a test as well, but that can come later

I have a patch to add the test, I will push it after this one.

sjoerdsimons:
@julien-massot can you post your test patch to this PR?

julien-massot:
> @julien-massot can you post your test patch to this PR?

@sjoerdsimons sure, it's posted.

cmd/fakemachine/main.go
```diff
@@ -18,6 +18,7 @@ type Options struct {
 	EnvironVars map[string]string `short:"e" long:"environ-var" description:"Environment variables (use -e VARIABLE:VALUE syntax)"`
 	Memory      int               `short:"m" long:"memory" description:"Amount of memory for the fakemachine in megabytes"`
 	CPUs        int               `short:"c" long:"cpus" description:"Number of CPUs for the fakemachine"`
+	SectorSize  int               `short:"S" long:"sectorize" description:"Override image sector size"`
```
obbardc, line 21:
```suggestion
	SectorSize  int               `short:"S" long:"sectorsize" description:"Override image sector size"`
```

machine.go
```diff
@@ -185,16 +185,17 @@ type image struct {
...
-	quiet    bool
-	Environ  []string
+	arch       Arch
+	backend    backend
+	mounts     []mountPoint
+	count      int
+	images     []image
+	memory     int
+	numcpus    int
+	sectorSize int
```
obbardc, line 195: This changes are not related

machine.go
```diff
@@ -496,6 +497,12 @@ func (m *Machine) SetNumCPUs(numcpus int) {
 	m.numcpus = numcpus
 }
 
+// SetSectorSize overrides the default sectors size(512) for the image exposed
```
obbardc, line 500:
```suggestion
// SetSectorSize overrides the default sector size(512 bytes) for the image exposed to the fakemachine
```

obbardc (changes requested):

cmd/fakemachine/main.go
julien-massot: Hi Chris, thanks, will fix

machine.go
obbardc: I think this is actually just from gofmt.

julien-massot: 'sectorSize' is longer than other struct members, so without this change, I'm hitting the linter error:

'''
machine.go:188:1: File is not properly formatted (gofmt)
	arch	  Arch
'''

sjoerdsimons (approved):

obbardc (approved):

Issue #237, opened 2025-08-08
trixie CI tests fail


obbardc:
CI test seems to fail:

```
=== NAME  TestScratchDisk
    machine_test.go:175: Test for device mount on scratch failed with -1
--- FAIL: TestScratchDisk (0.00s)
```

daissi:
I didn't have time and not sure I will have time for the next 2 weeks, so feel free to work on it.

obbardc:
I can't reproduce this by running the unit tests (e.g. ``) on my unstable system.

In the test-container I can reproduce the issue:

```

$ docker run \
  --rm \
  -it \
  -w /work \
  -v $(pwd):/work \
  --security-opt label=disable \
  --group-add=$(getent group kvm | cut -d : -f 3) \
  --device /dev/kvm \
  --cap-add=SYS_PTRACE \
  --tmpfs /scratch:exec \
  ghcr.io/go-debos/test-containers/trixie:main \
  go test ./... --backend=kvm

go: downloading github.com/klauspost/compress v1.17.9
go: downloading github.com/surma/gocpio v1.1.0
go: downloading github.com/docker/go-units v0.5.0
go: downloading github.com/ulikunitz/xz v0.5.12
go: downloading golang.org/x/sys v0.22.0
go: downloading github.com/jessevdk/go-flags v1.6.1
go: downloading al.essio.dev/pkg/shellescape v1.6.0
go: downloading github.com/stretchr/testify v1.10.0
go: downloading github.com/davecgh/go-spew v1.1.1
go: downloading github.com/pmezard/go-difflib v1.0.0
go: downloading gopkg.in/yaml.v3 v3.0.1
--- FAIL: TestScratchDisk (0.00s)
    machine_test.go:175: Test for device mount on scratch failed with -1
Running /a/b/c / using kvm backend
Running /fakemachine.test -test.run TestScratchTmp using kvm backend
Running /fakemachine.test -test.run TestImage512SectorSize kvm using kvm backend
Running 
MEM=$(grep MemTotal /proc/meminfo  | awk ' { print $2 } ' )
# MemTotal is usable ram, not physical ram so accept a range
if [ ${MEM} -lt 900000 -o ${MEM} -gt 1024000 ] ; then
  exit 1
fi
 using kvm backend
Running /fakemachine.test -test.v -test.run TestCommandEscaping -testarg '$s'"'"'n\akes' using kvm backend
[...]

PR #239, opened 2025-08-10
Test fakemachine in forky

obbardc:
Since forky is now the latest debian testing release, test fakemachine using the forky test containers.


obbardc:
Seems like unit tests fail due to same issue in #237 

```
```

PR #243, opened 2025-08-11
Create/update issue when scheduled CI job fails

obbardc:
Silly test for https://github.com/go-debos/debos/issues/573


obbardc:
rebased on #260

PR #244, opened 2025-08-11
Deprecate user-mode-linux backend

obbardc:
Remove user-mode-linux backend from automatic backend selection & allow user-mode-linux backend to be used manually along with a warning. Remove the user-mode-linux CI tests.


obbardc:
> Should we leave the UML tests that are current passing? Or are they already causing issues?

I can do that. It was just easier to remove them all :-).

PR #260, opened 2025-12-01
ci: remove bookworm tests

obbardc:
fakemachine is no longer supported on bookworm; since golang minimum version requirement in fakemachine has been bumped to 1.24. Remove the tests for fakemachine on bookworm.


.github/workflows/ci.yml
```diff
@@ -50,14 +50,12 @@ jobs:
         # See https://github.com/actions/runner-images/issues/183
         #
         # For Arch Linux uml is not yet supported, so only test under qemu there.
-        os: [bookworm]
-        backend: [qemu, uml, kvm]
```
sjoerdsimons: Looks like we now do qemu  & kvm on all os', so i guess we can simplify te below again by using the matrix rather then including everything by hand?

PR #282, opened 2026-02-16
ci: Add Ubuntu and Fedora tests && update container image paths

obbardc:
We now build test-containers for Ubuntu & Fedora. Run the fakemachine test suite on these OSs as well.

Depends on https://github.com/go-debos/test-containers/pull/57

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,9 @@ jobs:
         backend:
           - qemu
           - kvm
+        exclude:
+          - os: fedora-latest
+            backend: qemu
     name: Test ${{matrix.os}} with ${{matrix.backend}} backend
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -54,8 +54,15 @@ jobs:
         #
         # For Arch Linux uml is not yet supported, so only test under qemu there.
         # user-mode-linux is currently broken; see https://github.com/go-debos/fakemachine/issues/241
-        os: [arch, trixie, forky]
-        backend: [qemu, kvm]
+        os:
+          - arch
+          - trixie
+          - forky
+          - plucky
```
obbardc: this should be called `ubuntu-plucky` i guess

PR #297, opened 2026-04-19
fix broken udev symlinks for >26 disks and >9 partitions

obbardc:
Two very minor bugs in udev rule generation caused /dev/disk/by-fakemachine-label/ symlinks to silently not be created when:

- attaching more than 26 disks
- attaching a disk with more than 9 partitions

In practice these bugs will probably never be hit by normal users of fakemachine, but let's fix these silly bugs anyway.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -485,6 +485,16 @@ func (m *Machine) CreateImage(imagepath string, size int64) (string, error) {
 	return m.CreateImageWithLabel(imagepath, size, label)
 }
 
+// diskSuffix returns the disk name suffix for the i-th disk (0-indexed),
+// following Linux device naming: a, b, ..., z, aa, ab, ..., az, ba, ...
+func diskSuffix(i int) string {
+	suffix := ""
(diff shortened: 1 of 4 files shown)


machine.go
```diff
@@ -485,6 +485,18 @@ func (m *Machine) CreateImage(imagepath string, size int64) (string, error) {
...
 
+// diskSuffix returns the disk name suffix for the i-th disk (0-indexed),
+// following Linux device naming: a, b, ..., z, aa, ab, ..., az, ba, ...
+func diskSuffix(i int) string {
+	suffix := ""
+	for n := i + 1; n > 0; {
+		n--
+		suffix = string(rune('a'+n%26)) + suffix
+		n /= 26
+	}
```
sjoerdsimons:
```suggestion
	for ; i >= 0 ; i = (i/26 - 1) {
		suffix = string(rune('a'+i%26)) + suffix
	}
```

obbardc: done, thanks

PR #299, opened 2026-04-19
ci: re-enable fedora tests on qemu

### LINKED ISSUE #283: Fix unit tests on fedora with qemu
In https://github.com/go-debos/fakemachine/pull/282 we added a CI job to run the unit tests on fedora. As part of that, the unit tests seems to always fail with qemu on fedora, so they were disabled.

We need to find out the root cause, fix the unit tests on fedora/qemu & revert the commit which disables the unit tests on fedora/qemu.

obbardc:
We disabled the fedora tests as they were broken (strangely enough only on qemu). Re-enable the tests now they are stable.

Closes: #283

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,9 +60,6 @@ jobs:
         backend:
           - qemu
           - kvm
-        exclude:
-          - os: fedora-latest
-            backend: qemu
     name: Test ${{matrix.os}} with ${{matrix.backend}} backend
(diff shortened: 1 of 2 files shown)


obbardc:
still broken :(

obbardc:
removing `t.Parallel()` adds only a minute to the CI runtime (overall 5 minutes). Think it's good to go.

cc @sjoerdsimons

PR #301, opened 2026-04-21
machine: add SetScratchMkfsArgs to allow customising scratch mkfs options

obbardc:
When the scratch filesystem is backed by a disk image it is formatted with mkfs.ext4 using only default options. Some callers may need to adjust those options. For example, debos builds for 32-bit target architectures need to disable the dir_index feature.

This is needed by https://github.com/go-debos/debos/pull/704

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -197,11 +197,12 @@ type Machine struct {
 	quiet      bool
 	Environ    []string
 
-	scratchsize int64
-	scratchpath string
-	scratchfile string
-	scratchdev  string
-	initrdpath  string
+	scratchsize     int64
+	scratchpath     string
+	scratchfile     string
+	scratchdev      string
+	scratchMkfsArgs []string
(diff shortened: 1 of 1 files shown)


sjoerdsimons:
the scratch filesystem is only meant to be mounted by the vm for `/scratch` ; So i'm a bit confused by the rationale for this one. Tell me more :)

obbardc:
> the scratch filesystem is only meant to be mounted by the vm for `/scratch` ; So i'm a bit confused by the rationale for this one. Tell me more :)

@sjoerdsimons this was created to fix https://github.com/go-debos/debos/issues/620 along with https://github.com/go-debos/debos/pull/704

The rationale for this change should be in the linked PR and issue. Does it make sense ?

Happy to solve this in some other way if we can :-). Possibly better to continue the discussion on the linked PR.

PR #308, opened 2026-05-03
docs: clarify fakemachine runtime environment

### LINKED ISSUE #303: Docs: Clarify the intended use of fakemachine
So I know this project is to "fake a machine", and that is to

> Creates a virtual machine based on the currently running system.

But "currently running system" isn't clear to me.

Seeing this from the DebOS perspective, my assumption is that the currently running system is meant to be a container and you're sharing a small number of `/dev/kvm` host resources without needing root access, all while having a preconfigured build environment to run modifications to the VM and output the result. Is that accurate?

If so, would a description like this be appropriate?

> A QEMU/KVM backend API that runs privileged build tasks on a VM running inside container and CI environments without requiring root on the host. This combines the isolation and portability benefits of containers with fast access to hardware on the host.

If this is what it does, I would hope to create my own build and testing Debian containers to reap all the container benefits (consistent tooling, running on Ansible's molecule framework, running on any Posix OS, etc..).

obbardc:
Clarify what is meant by creating a VM from the "currently running system". The phrase refers to the environment fakemachine is invoked from: this may be a host system, container or CI runner.

Expand the README to describe that fakemachine is mainly used by debos to execute actions inside a VM derived from the current runtime environment. Also clarify that container usage is common but not required and that KVM is optional for acceleration rather than a hard dependency.

Closes: #303

### DIFF
--- a/doc/man/fakemachine.1
+++ b/doc/man/fakemachine.1
@@ -4,7 +4,21 @@
 .SH NAME
 fakemachine \- fake a machine
 .PP
-Creates a virtual machine based on the currently running system.
+fakemachine creates a virtual machine derived from the current root
+filesystem and allows the user to run commands inside it.
+.PP
+It is provided as a standalone executable as well as a Go library and is
+primarily used by \c
+.UR https://github.com/go-debos/debos
+debos
+.UE \c
+\ to execute image build actions inside a VM.
+This gives stronger isolation than chroot and allows unprivileged users
+to perform operations that would normally require root on the host, such
+as mounting filesystem images.
+.PP
+If fakemachine is ran inside a container, the virtual machine\(cqs root
+filesystem is derived from the container\(cqs root filesystem.
 .SH SYNOPSIS
 .IP
 .EX
--- a/README.md
+++ b/README.md
@@ -1,6 +1,16 @@
 # fakemachine - fake a machine
 
-Creates a virtual machine based on the currently running system.
+fakemachine creates a virtual machine derived from the current root filesystem
+and allows the user to run commands inside it.
+
+It is provided as a standalone executable as well as a Go library and is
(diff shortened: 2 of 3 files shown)


doc/man/fakemachine.1
```diff
@@ -4,7 +4,33 @@
...
-Creates a virtual machine based on the currently running system.
+fakemachine creates a virtual machine derived from the environment it is
+invoked from and allows the user to run commands inside the virtual
+machine.
+.PP
+It is provided as a standalone executable as well as a Go library.
+.PP
+The \(lqenvironment it is invoked from\(rq means the runtime environment
+where fakemachine is started.
+This can be a host system, a Docker/Podman container or a CI runner.
```
sjoerdsimons: what's  a CI runner in this case and how is it not either a container or a host system?

doc/man/fakemachine.1
```diff
@@ -4,7 +4,33 @@
...
+fakemachine uses that environment as the base for the VM rather than
+constructing a separate root filesystem.
+For example, when fakemachine is ran inside a Debian container, the VM
+will effectively be based on that container\(cqs userspace.
```
sjoerdsimons: this whole paragraph seems a bit confusing as it feels like too many different terms are used to vaguely refer to the same thing. (e.g. root fs, userspace, runtime environment)  and then mixing container, host system and ci runner.. Tbh. probably best to just focus on it re-using the current root file system (whether is a container rootfs or otherwise)

obbardc: done

doc/man/fakemachine.1
obbardc: done

bitsondatadev:
@obbardc sorry it took me a minute. This all reads clearly to me now. No notes! :)

PR #313, opened 2026-05-20
Make sure to create image file from scratch

richardweinberger:
Truncate to 0 while opening the file, otherwise when the image file already exists, old data will remain, making bmaptool inefficient. Beside of that it's a potential info leak.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -522,7 +522,7 @@ func (m *Machine) CreateImageWithLabel(path string, size int64, label string) (_
 		}
 	}
 
-	flags := os.O_WRONLY
+	flags := os.O_WRONLY | os.O_TRUNC
 	if size >= 0 {
 		flags |= os.O_CREATE
 	}


obbardc (approved):

richardweinberger:
Hmm. I'm not sure whether the failing tests are related to my change.

obbardc:
> Hmm. I'm not sure whether the failing tests are related to my change.

Don't think so from a quick look. Will look into it.

obbardc:
@richardweinberger seems like this PR is trying to push to the container registry from the PR context, which obviously you don't have write access to.

Can you try to rebase your patch on top of https://github.com/go-debos/fakemachine/pull/314 once the CI there passes ? I'd like to test that in the context of this PR.

obbardc:
> @richardweinberger seems like this PR is trying to push to the container registry from the PR context, which obviously you don't have write access to.
> 
> Can you try to rebase your patch on top of #314 once the CI there passes ? I'd like to test that in the context of this PR.

I've merged that, hopefully once you rebase on main your PR should pass CI :-)

PR #314, opened 2026-05-20
ci: avoid pushing container images from PR workflows

obbardc:
Pull request workflows may not have permission to write to the organisation's container registry (e.g. when triggered from forks or by users who are not members of the organisation).

Only push container images for non-PR events while still allowing PRs to build the image.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,7 +150,8 @@ jobs:
       uses: docker/build-push-action@v7
       with:
         context: .
-        push: true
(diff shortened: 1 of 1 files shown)


obbardc:
Need to check if a container actually gets pushed for this branch.

obbardc:
> Need to check if a container actually gets pushed for this branch.

Yep, the container gets pushed for the `push` event but the container for the `pull_request` event doesn't get pushed, which is a useful cleanup anyway.

PR #320, opened 2026-06-30
Revert "Make sure to create image file from scratch"

sjoerdsimons:
Revert #313  as it causes regression to fakemachine and violated the documented function contract.

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -522,7 +522,7 @@ func (m *Machine) CreateImageWithLabel(path string, size int64, label string) (_
 		}
 	}
 
-	flags := os.O_WRONLY | os.O_TRUNC
+	flags := os.O_WRONLY
 	if size >= 0 {
 		flags |= os.O_CREATE
 	}


sjoerdsimons:
Surprised this didn't break the tests tbh ; Should really add a test for this ;) cc: @richardweinberger

obbardc:
> Surprised this didn't break the tests tbh ; Should really add a test for this ;) cc: @richardweinberger

Added a test for this in https://github.com/go-debos/fakemachine/pull/323 ; fails in main and passes with this PR...

richardweinberger:
Please help me to understand this revert. What is the use case of having size being -1? Does debos even allow this?
I always thought the `imagesize` yaml setting is mandatory and a unit.

Even if there is a valid use case, it's still a foot gun to have the image not being truncated on repeated debos runs.
Stale data from old runs will be there. This is not only a problem for reproducible builds also harms bmaptool, as I wrote in the commit message.

obbardc:
> Please help me to understand this revert. What is the use case of having size being -1? Does debos even allow this? I always thought the `imagesize` yaml setting is mandatory and a unit.
> 
> Even if there is a valid use case, it's still a foot gun to have the image not being truncated on repeated debos runs. Stale data from old runs will be there. This is not only a problem for reproducible builds also harms bmaptool, as I wrote in the commit message.

@richardweinberger suggest you create a new issue to discuss this (linking this revert PR and your original PR); context will get lost in this closed PR :-).

